PR #593, opened 2026-06-05
fix(optimization): stop prompting default tool paths

chennyiiis:
## Summary
- Stop generated optimization prompts from defaulting OOB and TraceLens tool paths.
- Add a single-session safety constraint so recovery uses `optimize --resume` instead of launching a new session.

## Test plan
- `gofmt -w SaFE/apiserver/pkg/handlers/optimization/prompt_builder.go SaFE/apiserver/pkg/handlers/optimization/prompt_builder_test.go`
- `go test ./SaFE/apiserver/pkg/handlers/optimization` was attempted, but the local dependency graph fails before package tests with `k8s.io/apimachinery` / `structured-merge-diff` v4-v6 type mismatch.

### DIFF
--- a/SaFE/apiserver/pkg/handlers/optimization/handler.go
+++ b/SaFE/apiserver/pkg/handlers/optimization/handler.go
@@ -868,7 +868,6 @@ func promptConfigFromRequest(req *CreateTaskRequest, m *ResolvedModel, workspace
 		MaxHours:       req.MaxHours,
 		TargetGain:     req.TargetGain,
 		Image:          req.Image,
-		InferenceXPath: req.InferenceXPath,
 		OOBPath:        req.OOBPath,
 		TraceLensRoot:  req.TraceLensRoot,
 		Workspace:      workspace,
--- a/SaFE/apiserver/pkg/handlers/optimization/prompt_builder.go
+++ b/SaFE/apiserver/pkg/handlers/optimization/prompt_builder.go
@@ -15,25 +15,24 @@ import (
 // so that a task created via SaFE produces the same prompt as the same form
 // submitted through the Hyperloom UI.
 const (
-	defaultMode           = ModeClaw
-	defaultFramework      = FrameworkSGLang
-	defaultPrecision      = "FP4"
-	defaultGPUType        = "MI355X"
-	defaultISL            = 1024
-	defaultOSL            = 1024
-	defaultConcurrency    = 64
-	defaultTP             = 1
-	defaultEP             = 1
-	defaultInferenceXPath = "/hyperloom/InferenceX"
-	defaultResultsPath    = "/workspace/hyperloom/"
-	defaultGeakStepLimit  = 100
-	defaultMaxHours       = 3.0
-	defaultTargetGain     = 30.0
-	defaultRayReplica     = 1
-	defaultRayGpu         = 1
-	defaultRayCPU         = 12
-	defaultRayMemoryGi    = 128
-	raySharedMemoryGi     = 500
+	defaultMode          = ModeClaw
+	defaultFramework     = FrameworkSGLang
+	defaultPrecision     = "FP4"
+	defaultGPUType       = "MI355X"
+	defaultISL           = 1024
+	defaultOSL           = 1024
+	defaultConcurrency   = 64
+	defaultTP            = 1
+	defaultEP            = 1
+	defaultResultsPath   = "/workspace/hyperloom/"
+	defaultGeakStepLimit = 100
+	defaultMaxHours      = 3.0
+	defaultTargetGain    = 30.0
+	defaultRayReplica    = 1
+	defaultRayGpu        = 1
+	defaultRayCPU        = 12
+	defaultRayMemoryGi   = 128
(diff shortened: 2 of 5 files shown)


chennyiiis:
## Change Notes

This PR is the SaFE-side half of the Hyperloom optimization prompt cleanup. Full change set across both repos:

### Hyperloom-side changes

- Stopped forcing tool source paths from the Hyperloom submit client:
  - `inferencexPath` is no longer sent by default.
  - `oobPath` and `tracelensRoot` are no longer sent by default.
  - These fields are now only sent when explicitly provided by CLI flags or env vars.
  - Rationale: sandbox-side Hyperloom setup / `install.sh` should own tool detection, clone, and exported runtime paths.

- Simplified `ci/prompt_prefix.txt`:
  - Removed redundant `USER_DATA_PATH` export and directory creation.
  - Removed prompt-prefix audit file writes (`hyperloom_source_commit.txt`, `source_commits.json`).
  - Removed extra env exports such as `CORTEX_KB_URL` / `NODE_TLS_REJECT_UNAUTHORIZED` from the prompt prelude.
  - Kept the sandbox safety guardrails around broad `find`, unsafe `kill -9`, and per-bash timeout limits.
  - Added the explicit shell-prelude instruction at the top.

- Added the single-session runtime constraint to the Hyperloom prompt prefix locally:
  - After first launch, do not start another `optimize --model` / “Launch a New Optimization”.
  - Terminal `stop_reason` in `state.json` is final.
  - Crash recovery must use `optimize --resume` in the same session dir.

### SaFE-side changes in this PR

- Removed SaFE prompt-builder defaults for optional tool paths:
  - Removed `/hyperloom/OOB` defaulting.
  - Removed `/hyperloom/TraceLens-internal` defaulting.

- Stopped emitting these lines in generated optimization prompts:
  - `OOB path: ...`
  - `TraceLens path: ...`

- Kept the API fields for compatibility:
  - `oobPath`
  - `tracelensRoot`
  - They are no longer default-filled or printed into the prompt by SaFE.

- Added the same single-session safety requirement to the end of SaFE-generated prompts:
  - Never launch a second optimization session after the first launch.
  - Treat `state.json` `stop_reason` as final.
  - Use only `optimize --resume` for unexpected crash recovery.

- Updated prompt-builder tests:
  - Assert OOB / TraceLens path lines are absent.
  - Assert the single-session safety text is present for both claw and local prompt modes.

### Validation

- Ran `gofmt` on the touched Go files.
- Attempted `go test ./SaFE/apiserver/pkg/handlers/optimization` locally, but it fails before package tests due to the existing dependency mismatch:
  - `k8s.io/apimachinery/pkg/util/managedfields/internal/typeconverter.go`
  - `structured-merge-diff` v6 `TypeDef` cannot be used as v4 `TypeDef`.
- This failure is in dependency compilation and is not caused by the prompt-builder assertions changed here.

### Notes

- InferenceX is intentionally left as-is on the SaFE side in this PR only where existing generated prompt still says `InferenceX path`; the immediate cleanup here targets OOB and TraceLens defaults that were still being injected by SaFE even after Hyperloom stopped sending them.
- Hyperloom's submit client now makes these tool path overrides explicit opt-in, while this PR prevents SaFE from silently reintroducing default OOB / TraceLens paths when the client sends empty values.

chennyiiis:
## Cross-Repo Change Summary

This SaFE PR is paired with Hyperloom PR AMD-AGI/Hyperloom#452.

SaFE-side changes in this PR:

- Stop `BuildHyperloomPrompt` from silently reintroducing default optional tool paths when the client leaves them empty.
- Remove default prompt-builder values for:
  - `/hyperloom/OOB`
  - `/hyperloom/TraceLens-internal`
- Stop emitting these generated prompt lines:
  - `OOB path: ...`
  - `TraceLens path: ...`
- Keep `oobPath` and `tracelensRoot` request fields for API compatibility; SaFE just no longer defaults or prints them.
- Append the single-session safety rule to generated optimization prompts:
  - do not start a second `optimize --model` / “Launch a New Optimization” after the first launch;
  - terminal `state.json` `stop_reason` is final;
  - unexpected crash recovery must use `optimize --resume` in the same session dir.
- Updated prompt-builder tests to assert OOB/TraceLens path lines are absent and the single-session rule is present.

Hyperloom-side paired changes:

- Hyperloom submit now treats InferenceX/OOB/TraceLens paths as explicit opt-in overrides instead of default SaFE fields.
- Hyperloom prompt prefix was simplified to rely on SaFE-provided `USER_DATA_PATH` and Hyperloom runtime session manifests, rather than writing path/audit setup in the prompt prefix.
- This keeps tool setup ownership on the sandbox-side Hyperloom runtime / `install.sh`, while this SaFE PR prevents backend prompt defaults from reintroducing stale paths.

Validation notes:

- `gofmt` was run on the touched Go files.
- `go test ./SaFE/apiserver/pkg/handlers/optimization` was attempted locally, but dependency compilation fails before package tests due to an existing `k8s.io/apimachinery` / `structured-merge-diff` v4-v6 type mismatch.

weilei0120 (approved):

PR #605, opened 2026-06-16
add kgateway

weilei0120:


chaojhou (commented):
AddonTemplates look good. Before merge: (1) split or flag the `install.sh` fix and add a shell-lint gate, (2) enforce/loudly document CRD-before-controller ordering, (3) verify data-plane images actually resolve to Harbor on an air-gapped install.

chaojhou (approved):

Issue #588, opened 2026-06-02
validateOwnerWorkload admission check rejects multi-node CICD UnifiedJobs (preheat child created before its owner)


amd-ama10002-2:
## Summary

Multi-node CICD jobs (e.g. `ROCm/unified-training-dockers` `multi_node_tests` on the `utd-multi-node-qqcws` runner set) stopped working: the unified training job is never created. The `cicd-unified-job-proxy` launcher's `POST /api/v1/workloads` is rejected by the validating webhook with HTTP 400, the launcher writes `Failed` to SAFE_OUTPUT, and the GitHub job fails within ~5s without any training running.

Root cause is the `validateOwnerWorkload` admission check added in #559 ("supports workload owner", 2026-05-12), combined with the apiserver creating **preheat child workloads before their owner workload exists**.

## Impact

- All multi-node unified CICD jobs fail at admission time (single-node / build jobs are unaffected because they don't create an owner-linked child).
- Regression introduced by #559; before that the owner label was an unenforced cleanup hint.

## Root cause

`createWorkload` creates the preheat children **before** the main workload:

`SaFE/apiserver/pkg/handlers/resources/workload.go`
```go
if req.Preheat {
    preheatWorkloads, err = h.createPreheatWorkloads(c, mainWorkload, req, requestUser, roles) // created first
    ...
}
resp, err := h.createWorkloadImpl(c, mainWorkload, requestUser, roles)                          // main created after
```

Each preheat child is labeled with its owner = the (not-yet-created) main workload:
```go
Labels: map[string]string{
    v1.DisplayNameLabel: displayName,
    v1.OwnerLabel:       mainWorkload.Name,   // primus-safe.owner -> main UnifiedJob
},
```

The webhook then rejects the child because the owner doesn't exist yet:

`SaFE/webhooks/pkg/workload_webhook.go`
```go
func (v *WorkloadValidator) validateOwnerWorkload(ctx context.Context, workload *v1.Workload) error {
    ownerId := workload.GetOwnerWorkloadId()
    if ownerId == "" {
        return nil
    }
    ...
    owner := &v1.Workload{}
    if err := v.Client.Get(ctx, client.ObjectKey{Name: ownerId}, owner); err != nil {
        return commonerrors.NewBadRequest(
            fmt.Sprintf("ownerWorkloadId %q not found: %v", ownerId, err))
    }
    ...
}
```

## Evidence

[...]

PR #611, opened 2026-06-18
test(webhooks): align owner-workload tests with NotFound tolerance

zili-amd:
validateOwnerWorkload now tolerates a NotFound owner workload (issue #588): an owner-labeled child can be admitted before its owner is persisted, so a missing owner must not block admission. Update the stale assertions that still expected an error for a missing owner to expect nil, and keep error-branch coverage via a self-referential owner (which is still rejected).


amd-ama10002-2 (approved):
I'm approving thanks! Might also be a good idea to run this by @weilei0120 if he has the time because he understands the workload scheduling better than I do 😅

weilei0120 (approved):

PR #613, opened 2026-06-19
fix(docs): sync docs-site lockfile with package.json for npm ci

zili-amd:
The Pages deploy workflow runs `npm ci`, which requires package.json and package-lock.json to be in sync. They had drifted (package.json pinned @docusaurus/* 3.6.3 while the lockfile resolved 3.10.1), so `npm ci` aborted with EUSAGE. This pins @docusaurus/* to 3.10.1 and regenerates the lockfile. Verified `npm ci` + `npm run build` locally.

Supersedes #612 (closed when its branch was renamed to the `fix/zicanli/XXX` convention).

### DIFF
--- a/docs-site/package-lock.json
+++ b/docs-site/package-lock.json
@@ -8,9 +8,9 @@
       "name": "primus-safe-docs",
       "version": "0.0.0",
       "dependencies": {
-        "@docusaurus/core": "^3.10.1",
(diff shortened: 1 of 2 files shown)


amd-ama10002-2 (approved):
LGTM! 🙂👍

PR #625, opened 2026-06-26
docs: add docs-as-test harness and polish GA install/tasks/admin pages

zili-amd:
Treat the docs-site as the test suite: pages carry reader-invisible  annotations (verify/behavior/contract) that an agent executes against a live console via a browser MCP. A build-time remark plugin strips the annotations from the shipped HTML so they stay source-only.

- Add docs-site/AGENTS.md (agent contract, taxonomy, coverage map) and docs-site/RUN-DOCS-AS-TEST.md (copy-paste run trigger); runtime params live in a gitignored .docs-test.env (with .example template) so no host/credentials are committed.
- Install: accurate paths, hosts.ini example, storage/higress/harbor/bench steps, required-vs-optional, NIC params; trim internal info.
- Tasks: console-first single-node + new multi-node page; merge Authoring into "Develop & interact with your jobs"; add console screenshots.
- Administration: console emphasis + tests for users (RBAC), access & quota (API key), and nodes (read-only inventory).

## Docs-as-test: known coverage gaps (follow-up, addressed separately)

**Behavior tests deferred — destructive / need a disposable cluster**
- `administration/manage-nodes` — mutating node ops (taint/cordon/drain, reboot, delete) are destructive on a live cluster; test only against a disposable node.
- `administration/upgrading` — System → Deploy create/approve/rollback are destructive; test on a disposable env (presence check is Done).
- `getting-started/first-training-job` — capacity onboarding (SSH secret, nodes, cluster, workspace) needs a disposable cluster to register against.

**Behavior tests deferred — need external fixtures / build flags**
- `tasks/run-cicd-runners` — needs the ARC add-on + a GitHub App/PAT + a repo, and a CICD-visible build (the console page is hidden in GA-scoped builds).
- `tasks/interact-with-your-job` — terminal SSH (port 2222) needs an SSH client outside the browser, so it can't be part of the UI-only run.

**Docs need more detail before a precise test**
- `administration/manage-users` — state the exact non-admin behavior on System pages (hidden nav vs redirect); add the positive path (grant a workspace, confirm the user can act in it).
- `getting-started/prerequisites` — add minimum CPU/memory sizing per control-plane scale (small/medium/large); document expected RDMA NIC naming.
- `tasks/interact-with-your-job` — document the Save Image flow, the Resume-a-stopped-box flow, and where results go (workspace PFS checkpoints/outputs, dump-log/download, S3 export).

**Concept/overview pages without a behavior test yet (verify-only TODO)**
- `concepts/workload-types` (assert named artifacts only), `concepts/fault-tolerance`, `concepts/storage-and-data`, `architecture`, `intro`.


docs-site/docs/administration/manage-nodes.md
amd-ama10002-2, line 179: Might be good to explain how to do this node movement in the UI by clicking buttons. You could explain this using screenshots, or you could just say what buttons to click like this: https://tas325.primus-safe.amd.com/login-admin

It might be good to give a little more detail about what side panel or tab they need to be in in order to do this, because that's how we most often move notes around.

amd-ama10002-2, line 190: I guess you're planning to add a screenshot for this, and I think that would be helpful if you have one 🙂

amd-ama10002-2, line 280: We should probably give instructions about how to reboot the node by clicking buttons in the UI because I think that's the more common way we reboot nodes. 

Unless maybe this is just the docs for only using REST API commands? And the UI guide is one of these other doc files 🙂

amd-ama10002-2: We should probably also give some UI guide, some guide on how to use the UI to do this removal of a node from the cluster. Also, it's worth mentioning it takes like ten to fifteen minutes for the node to fully unmanage, and that's expected.

amd-ama10002-2, line 322: Yes, this is a good note. I think we should give some explanation about how to make a node template and how to add a node via the UI

amd-ama10002-2, line 55: I think there is an additional step here where you need to take the SSH public key that exists on the deploy note, and you need to put that under the root user's authorized keys on each node that you want to add into the cluster. That's a really important step, actually 😅  I'm not sure if maybe you're planning to add this later or maybe you have this information somewhere I overlooked it

docs-site/docs/administration/manage-users.md
amd-ama10002-2, line 69: I like your instructions on how to navigate the UI. Nice job! I feel like this step-by-step instruction on what buttons to click is super handy, and maybe we should try to put this kind of step-by-step instruction in some of the other sections too

amd-ama10002-2: I'm not sure if we need to go into further detail. Maybe you could save yourself a little work here. It's probably enough for most people to just say that the username is Groove, the password is Groove, and then I guess if users want to figure out where the default password is set, they can use AI to scan the repo or something.

docs-site/docs/administration/observability.md
amd-ama10002-2, line 1: This looks correct, but as you know, we should probably hold off on merging this until after we confirm the observability actually works.  

 This doc might be a good candidate for putting on a separate branch. If your plan is to merge this existing PR and then update the docs with a follow-up PR

docs-site/docs/administration/observability.md
```diff
@@ -0,0 +1,93 @@
...
+## Validate hardware with Primus-Bench
+
+Primus-Bench is separate from live monitoring: it validates node **health and performance** rather
+than watching a running job. Two ways to use it:
+
+- **From the console** — **System → Bench → Create Bench** runs a check against a node, cluster, or
+  workspace and reports pass/fail. Full steps are in
+  [Pre-flight & in-flight monitoring](/administration/preflight-and-monitoring).
+- **Standalone** — run the module directly (bare-metal / SLURM / Kubernetes) for deeper I/O and
+  system benchmarks, per [`Bench/README.md`](https://github.com/AMD-AGI/Primus-SaFE/blob/main/Bench/README.md).
```
amd-ama10002-2: Not sure why we have this in observability. We don't need this section in this doc, right? 🤔🤷‍♂️

docs-site/docs/administration/preflight-and-monitoring.md
```diff
@@ -5,83 +5,78 @@ title: Pre-flight & in-flight monitoring
...
+
+In the console, go to **System → Bench** and click **Create Bench**:
+
+1. Set a **name**, an **image**, and the **entry point** (the check to run).
+2. Choose the **Type** (node / cluster / workspace) and the **Value** (the specific target), and
+   set the **resources** for the check.
+3. Optionally enable **Toleration** so the check can run on nodes that are already tainted, and set
+   a **Timeout**.
+4. **Submit**, then track it in the list — its **Phase** moves to `Succeeded` or `Failed`. Open the
+   job to read its report.
```
amd-ama10002-2, line 25: I'm not sure if we have a good example of a Docker image to use for the pre-flight checks. 

I think that Wei, Lei has had to build a new Docker image for every new cluster we were trying to do pre-flight checks for. So, I think maybe each new cluster requires a custom Docker image, which I guess is not super helpful for a preflight check 😅

Because of this, it might be better to just remove the preflight checks from this doc. 

If we want to keep the preflight checks docs, we would need to either give some instructions on how to build the Docker image for a new cluster (which could be a lot of work), or we try to maybe suggest using the pre-flight checks of the Primus CLI.

docs-site/docs/concepts/fault-tolerance.md
amd-ama10002-2, line 63: I think you could put node operations in either document, but I have a slight preference for a document called operations if you create one.  Also, I don't think the UI has the ability to cordon or drain or export a node, as far as I know, so I'm not sure if we need to cover that?

docs-site/docs/concepts/storage-and-data.md
amd-ama10002-2: I think you did a good job with explaining the storage. There's a lot of detail here already. I don't know if you need to go into these additional details, at least for this first round of documentation. Maybe this can be, in the future release, that we add more details.

If we need to give an exact mount path, I think that's something that's configured when creating a workspace. I think it's fairly intuitive. The UI makes it pretty clear, so I don't know if it needs much explanation other than just showing the user that we have this feature.

I'm not sure what the PV template details are. I also think size and guidance for PFS capacity versus node local capacity. I think that's hard for us to advise on because that's very dependent on what the user's cluster looks like. I think we can save ourselves the work and just not write this if that's okay with you.

docs-site/docs/concepts/workload-types.md
amd-ama10002-2, line 19: I'm looking at what's deployed on TAS 300. Under inference workloads, I don't see a StatefulSet, so I'm wondering if maybe these inference workloads need to be updated. Deployment looks right, though.

docs-site/docs/getting-started/install.md
```diff
@@ -5,167 +5,219 @@ title: Install
...
+node-01
+node-02
+node-03
+
+[kube_node]
+node-01
+node-02
+node-03
+node-04
+# ...and every other GPU worker node
```
amd-ama10002-2: Actually, the recommended practice is to not add any extra nodes for the workers. Here's a note I have from AI 

> Do not add any EXTRA nodes under the kubegroup in hosts.ini.  
> The recommended practice is to:
> 
> Use this file hosts.inisolely to define and configure the control plane and etcd nodes.
> 
> The managing of new worker nodes is intended to be done through the Primus-Safe UI.
> 
> Using the kube_node group makes it so that these control plane nodes can run workloads too

docs-site/docs/getting-started/install.md
```diff
@@ -5,167 +5,219 @@ title: Install
...
-## Single cluster (the common path)
+Here `node-01`–`node-03` are the control-plane / etcd members, and every node (including more
+workers beyond `node-04`) goes under `[kube_node]` so it can run GPU workloads.
 
-### 1. (Bare metal only) Build the cluster with Bootstrap
+- `ansible_host` / `ip` are the addresses the installer and the other nodes use to reach each
+  machine — use the private cluster network.
+- Put every GPU worker in `[kube_node]`. A node may be both a control-plane node and a worker.
+- If the deploy host reaches the nodes with a non-default SSH key, add
+  `ansible_ssh_private_key_file=/path/to/key` to each `[all]` line.
```
amd-ama10002-2: Also worth noting here that running the Kubespray will rename the host names of the nodes to what you specified in this host.ini file. At least that's what I think it does, if I remember correctly. 

This is worth mentioning because the hostname of the node is needed to stay the same so that it can work with something like IAM, Identity Management.

docs-site/docs/getting-started/install.md
```diff
@@ -5,167 +5,219 @@ title: Install
...
+|--------|------------|
+| `ethernet nic` | The Ethernet interface distributed jobs use for NCCL/RCCL control traffic and TCP fallback (sets `NCCL_SOCKET_IFNAME`). Default `eno0`. |
+| `rdma nic` | The RDMA/RoCE devices distributed jobs use for high-speed GPU-to-GPU transfers (sets `NCCL_IB_HCA`). Default `rdma0,…,rdma7`. |
+| `storage_class` | The StorageClass from step 2 (default `local-path`; must already exist). |
+| `ingress` | `nginx` (console on NodePort `30183`) or `higress` (requires step 3). |
+| Image pull secret | Credentials for a private registry, or an empty placeholder. |
+| `cluster_scale` | `small` / `medium` / `large` — sizes the control-plane replicas and resources. |
+| S3 storage | Optional — endpoint, bucket, and keys for log download and object features. |
+| SSO / OIDC | Optional — connect an external identity provider. |
+| `csi_volume_handle` | Optional — enables a workspace persistent filesystem (PFS). |
```
amd-ama10002-2, line 197: Good. I thought there was also something where it asks for a cluster name, like this.  
```
Enter cluster name | tas325 | Becomes subdomain: tas325.primus-safe.amd.com
```
  I think this might be worth mentioning because we need it for accessing the cluster ingress later in this document.

docs-site/docs/getting-started/prerequisites.md
```diff
@@ -32,8 +31,7 @@ cluster's API):
   cluster add-on.
 - **High-speed networking for multi-node jobs** — RDMA / InfiniBand interfaces if you intend
   to run distributed training across nodes.
-- **Shared filesystem (optional)** — a CSI volume for workspace persistent storage (PFS). You
-  can enable this at install time or leave it disabled to start.
+- **Shared filesystem (optional)** — a high-performance Parallel File System (PFS) to support a production system.
```
amd-ama10002-2, line 34: Also, the networking on the cluster needs the following ports to be open: 

> Need ports 2222, 6443, 80, 443, 22 open
- 
- 2222 is used for the WebShell.
- 6443 is for the Kubernetes API server (optional I guess) 
- 443 is for https connections
- 80: I forgot what 80 is for but I think it's needed for something. 
- 22 is for ssh 

 Also, 2379 and 2380 are used for etcd.  It is not necessary to expose this outside the cluster, but inside the cluster, etcd needs to be able to communicate with each other over this port. Also, port 22 isn't necessary for exposing outside the cluster, but it is necessary, as you know, for ssh within the cluster.

docs-site/docs/tasks/interact-with-your-job.md
```diff
@@ -1,73 +1,114 @@
...
-The response lists each pod's `phase`, `nodeName`, and a ready-to-use `sshCommand`, plus
-`conditions` that explain scheduling/dispatch decisions.
+Your workspace must include the **Authoring** scope (ask a workspace admin if you don't see it —
+see [Workspace](/concepts/workspace)).
 
-## Logs
+1. Go to **Workloads → Authoring** and click **Create Authoring**.
+2. Set a **name**, an **image** (**Select** a registered image or **Custom** to type any pullable
+   reference), and the **resources** (CPU, GPU, memory, ephemeral storage). A dev box is always a
+   single node, so `replica` is fixed at 1.
```
amd-ama10002-2: This is optional, but maybe we should include information on how to use harbor and how to import a name into harbor. I am wondering if that is important because I guess a lot of people use that feature. 

 You can let me know if you think this is helpful or if this, adding a guide on how to import images with harvr, is too much information. 

Maybe this is not worth mentioning, but I think by default, if you don't specify with the prefix what registry you are pulling from, it will assume that you are pulling from docker.io. I guess you can also pull images from Quay or other OCI compatible registries.

amd-ama10002-2 (changes requested):
I'm doing two batches of reviews. 

This is just part one, and I'm going to have to do the other half of the files tomorrow. 

Overall, I think you did a really great job with this 🥳. I'm requesting a change because I think there are some missing pre-reqs that are actual blocking requirements for the cluster install, and then I think everything else is more like suggestions for your consideration, but not anything that should be blocking from getting merged or being released. 

I think that the docs are really detailed, and generally very good. For your consideration in future releases, I have a slight preference for a little less documentation. It's just my preference, but I think that providing less information can be helpful because it's less intimidating for readers and it helps readers find the information faster.

I think in the future, if you're open to it, we could consider removing content from some of these docs. For example, using the API is not something that is often done even by admins. We could remove references to the API, or make all of the API stuff one single doc, or tell users to use AI to read and learn about the API, then that could maybe cut back on a large portion of information.

Also, some of the UI features, I think, are pretty self-explanatory, so we don't need to give a whole lot of explanation of features beyond stating that a feature exists in the UI. I'm thinking, for example, with scheduling and authoring, maybe it's not important to explain the different priorities (low, medium, high). It's enough to just say that there is a priority that you can select from the dropdown, and only admins can use high priority. Stuff like this could help reduce the amount of content. Just a suggestion 😅👍

Again, nice work overall! I'll get to the second half of my PR review tmrw!

docs-site/docs/tasks/run-cicd-runners.md
```diff
@@ -0,0 +1,131 @@
...
+# The runner refuses to run as root; use an unprivileged user.
+RUN useradd --create-home --shell /bin/bash runner
+WORKDIR /actions-runner
+RUN curl -sL -o actions-runner.tar.gz \
+      "https://github.com/actions/runner/releases/download/v${RUNNER_VERSION}/actions-runner-linux-x64-${RUNNER_VERSION}.tar.gz" \
+ && tar xzf actions-runner.tar.gz && rm actions-runner.tar.gz \
+ && ./bin/installdependencies.sh && chown -R runner:runner /actions-runner
+USER runner
+CMD ["bash"]
+```
```
amd-ama10002-2, line 49: I think this example looks alright. I haven't tried it, but it looks like it will probably work. 

I want to point out that there exists images for GitHub Actions Runners that  are maintained by a third party: 
https://github.com/actions/runner/pkgs/container/actions-runner

and then you can look at this example of how we actually run a CI/CD: 
https://github.com/AMD-AGI/constellation/blob/1abecedbdd710fc125ecaf54c33662ef23108ebd/github_actions/Dockerfile#L1

docs-site/docs/tasks/run-cicd-runners.md
```diff
@@ -0,0 +1,131 @@
...
+    },
+    "githubAuth": { "type": "pat", "token": "<github-pat>" }
+  }'
+```
+
+Rotate credentials later with `PATCH /api/v1/workloads/<workloadId>` (send just `githubAuth`).
+
+> **Not yet covered (capture so we don't lose it):**
+> - [ ] Console screenshots of the CI/CD create wizard (Clone/Create + Advanced Options).
+> - [ ] Step-by-step for installing the ARC (`gha-runner-scale-set`) add-on from System → Addons.
```
amd-ama10002-2: I can't remember how we installed it. I don't remember going through the UI on the system panel. I'm thinking maybe this was one of the environment variable flags that I set when I did the bootstrap of save, but I'm not 100% sure. I guess you might not need to add this section because I'm not sure if this is how we often install it.

docs-site/docs/tasks/run-cicd-runners.md
```diff
@@ -0,0 +1,131 @@
...
+> - [ ] Where to watch workflow-run history (today it lives in the Lens app, not the SaFE console).
```
amd-ama10002-2: I think we often watch the workflow runs either in GitHub or we find it in the CI/CD tab on the side panel. You probably don't need to add this section of going to the workflow of the Lens app.   Maybe just adding a screenshot of the CI/CD in the UI will be enough.   

For example, there is a drop-down menu showing all the runs in the Primus-Safe UI. Taking a screenshot of just that is probably enough.

docs-site/docs/tasks/run-multi-node-training.md
```diff
@@ -5,30 +5,76 @@ title: Run a multi-node distributed job
...
+By default the scheduler chooses nodes. To target particular machines, use **Specified nodes**
+with a **node affinity** policy:
+
+- **`required`** — the job runs *only* on the listed nodes (it stays pending until they're free).
+- **`preferred`** — the scheduler prefers the listed nodes but may use others.
+
+When you specify nodes, the replica counts follow automatically (for PyTorchJob: master = 1,
+workers = the remaining nodes).
+
+## Networking for collectives
```
amd-ama10002-2, line 37: The content is generally good. 

I think maybe we could add a small note on the side. You need to use, sometimes it's required to use a docker image that has the correct RDMA drivers installed. For example, if we use Broadcom, people need to bring a docker that either has the Broadcom drivers installed or they need to run like some kind of fast script that installs the Broadcom drivers, that's their entry point. Similarly, for like AI-NIC usage, you need to install AI-NIC drivers. This is a common problem that people often forget to do. 

Additionally as a note, It can be helpful to use the pre-heat feature when starting a training job. I's uncommon, but a small number of users have met an issue where they try to start a large training job, like 32 node training job. Ine of the docker images will get pulled to a node very quickly. And then the rest of them are pulling very slow, and so the first docker image gets timed out before the other ones get pulled, causing the job to fail. 

So, using the toggle in the UI to preheat the image sometimes fixes this. It is an uncommon problem, but if the internet speed in the data center is not good for some reason, then this can be an issue. And it comes up more frequently if you use larger multi-nut training chops like 32, 64, 128N

docs-site/docs/tasks/run-single-node-training.md
```diff
@@ -78,15 +98,13 @@ Key fields:
...
-The full field list lives in the workload API (`SaFE/docs/apis/workload.md`).
-
 ## Your image
 
 Your job runs the container `image` you specify. You can:
 
 - Use a **public image** (e.g. an upstream PyTorch/ROCm image), or
-- Push to your **registry** and reference it, attaching a **registry secret** for private
-  images. See [Manage access & quota → Registry secret](/administration/manage-access-and-quota).
+- Push to your **registry** and reference it, attaching a **registry secret** for private images.
```
amd-ama10002-2, line 106: Might also be worth mentioning that users can import their image to the self-hosted harbor under the "Images" tab of the side-panel. Doing this import makes it slightly faster pulling an image instead of pulling over the open internet.

amd-ama10002-2 (commented):
Alright, this is part two of two of my PR review. I've looked at every document in here, and I think, in general, it's very good. I added a few comments for your consideration, but no hard blocker.

Just to repeat what I commented on the previous part one of my PR review: I think that these docs are really good work, and we can make a few small changes and release this as is. In the future, it might be worth considering trying to find a way to reduce the amount of information in these docs 🤷‍♂️🙂

docs-site/docs/administration/manage-nodes.md
amd-ama10002-2: Hi @zili-amd, I think it's important to mention this additional step somewhere. I'm not sure if you mentioned it somewhere else -- I don't think it's mentioned elsewhere, but maybe I overlooked it 😅

docs-site/docs/getting-started/install.md
amd-ama10002-2: @zili-amd I think this might be worth adding because the setting for the name is how the domain name gets created.

docs-site/docs/getting-started/prerequisites.md
amd-ama10002-2: @zili-amd This is the port stuff requirements that I think we should add. Having 2222 and 443 is needed for webshell and ssh access.

docs-site/docs/administration/manage-nodes.md
zili-amd: Added it as a warning right under Register a node in manage-nodes.md

docs-site/docs/getting-started/install.md
zili-amd: there's now a cluster name row at the top of the installer prompts table in install.md.  noting it becomes the subdomain (e.g. tas325 → tas325.primus-safe.amd.com) used to reach the cluster ingress later.

docs-site/docs/getting-started/prerequisites.md
zili-amd: Added a dedicated Network ports table in prerequisites.md

amd-ama10002-2 (approved):
I'm approving 🥳. 

I think the key comments that could potentially block cluster installation have been addressed, and the rest of the comments are things that can be resolved in follow-up PRs if we want. 

Thanks very much for this great work. I really appreciate your thorough documentation.

PR #648, opened 2026-07-06
fix(web): guard System admin pages with a system-admin route meta

### LINKED ISSUE #624: RBAC: default (non-admin) user can reach System admin pages (Nodes/Clusters) via direct URL
## Summary

A **`default`** (regular, non-admin) user is correctly prevented from seeing the **System** admin pages in the left-nav menu, but the pages are **still reachable by typing the URL directly**. Navigating to `/clusters` renders the full **Clusters admin page** — including the real cluster, a **Create Cluster** button, the **Is Protected** toggle, and per-row action buttons (edit/delete). `/nodes` likewise loads the admin Nodes view.

Access control here is enforced only by *hiding nav items client-side*; there is **no route-level (or, it appears, server-level) authorization** for these system-admin pages. This is an access-control / privilege-escalation gap.

## Environment

- Console: Primus-SaFE Web (`Web/apps/safe`), `v1.0.0`
- Reproduced against a live test env; also confirmed by code inspection on `main`.

## Steps to reproduce

1. As `system-admin` (e.g. `root`), create a local user (System → Users → Create) with role `default` and **no** workspace/role grants.
2. Sign out and sign in as that `default` user.
3. Observe the nav: the **System** admin pages (Clusters, Users, etc.) are hidden — good.
4. In the address bar, navigate directly to **`/clusters`**.
5. Navigate directly to **`/nodes`**.

## Expected

A `default` user has no system-admin privileges, so direct navigation to `/clusters`, `/nodes`, and other System pages should be denied — redirect to a 403 / home, the same way the nav hides them. The backend APIs backing these pages should also reject the request.

## Actual

- **`/clusters`** renders the **full Clusters admin page** for the `default` user: the real cluster is listed (Ready), with a **Create Cluster** button, an **Is Protected** toggle, and row action buttons. Because the page lists a real cluster, the backing **API returned admin data to a non-admin user**.
- **`/nodes`** loads the admin Nodes page.
- (For comparison, **`/usermanage`** does redirect away — so the gap is specifically the routes that lack a guard.)

> Note: write actions (Create/Delete/toggle protection) were **not** exercised, so it is not yet confirmed whether the backend would accept a mutating call from a `default` user. If it does, this should be escalated to **Critical / P0**. See open questions.

## Root cause (frontend)

There is no system-admin route guard. The relevant routes have **no `meta` guard at all**:

```45:59:Web/apps/safe/src/router/index.ts
        {
          path: '/nodes',
          name: 'Nodes',
          component: () => import('../pages/Nodes/index.vue'),
        },
        {
          path: '/nodedetail',
          name: 'NodeDetail',
          component: () => import('../pages/Nodes/NodeDetail.vue'),
        },
        {
          path: '/clusters',
          name: 'Clusters',
          component: () => import('../pages/Clusters/index.vue'),
        },
```

The auth guard only checks *authentication*, then lets any logged-in user proceed to any route:

```206:212:Web/apps/safe/src/router/guards/auth.ts
    // D) Already logged in — skip login-related pages
    if (user.isLogin && user.userId) {
      if (PUBLIC_NAME.has(to.name as string)) {
        return next((to.query.redirect as string) || '/')
      }
      return next()
    }
```

The only role check that exists is `requiresWorkspaceAdmin` (a *workspace*-admin check, not system-admin), and it is applied to the cluster **detail** route, not the cluster **list**:

```48:54:Web/apps/safe/src/router/guards/clusterReady.ts
      // Permission check (workspace-admin)
      if (to.meta?.requiresWorkspaceAdmin) {
        if (!wsStore.isCurrentWorkspaceAdmin() && !userStore.hasManagerAccess) {
          // TODO: redirect to 403 page instead of homepage
          return next('/')
        }
      }
```

So system-admin pages are only protected by being omitted from the nav menu — a client-side concern that direct URL navigation bypasses.

## Impact

- A regular user can view cluster/node administrative data and is presented with destructive admin controls (Create Cluster, delete, toggle cluster protection).
- Confirms the **backend returns admin resources to non-admin users** for at least the clusters listing — this is the more serious half and should be verified across all System endpoints.

## Suggested fix

1. **Backend (primary):** ensure all system-admin endpoints (clusters, nodes, etc.) enforce `system-admin` / `system-admin-readonly` authorization server-side and reject `default` users with 403. This is the real fix; the UI guard is defense-in-depth.
2. **Frontend:** add a `requiresSystemAdmin` route `meta` and a guard that redirects non-admins to a 403/home, applied to all System pages (`/clusters`, `/nodes`, `/nodedetail`, Flavors, Workspaces, Users, Faults, Registries, Addons, Deploy, Audit Logs, etc.). Mirror the existing `requiresWorkspaceAdmin` pattern in `clusterReady.ts`, or add a dedicated RBAC guard.
3. Consider a real **403 page** instead of redirecting to `/` (there is already a `// TODO: redirect to 403 page` note in the guard).

## Open questions

- Do the system-admin **mutating** APIs (POST/PATCH/DELETE on clusters/nodes) reject `default` users? If not → **Critical / P0**.
- Full inventory of System routes that lack a guard (the two confirmed are `/clusters` and `/nodes`).

---
_Found via the docs-as-test suite (`docs-site/AGENTS.md`) while executing the `administration/manage-users` contract: doc says "as that default user … Nodes/Clusters/Users not reachable"; product allows Nodes/Clusters via direct URL._

zili-amd:
System admin pages were hidden from the nav for non-admins but remained reachable by direct URL (e.g. /clusters, /deploy, /auditlogs), because the only route guards were authentication and a workspace-admin check. This is an access-control gap: a default user could open admin views directly.

- Add a `requiresSystemAdmin` route meta and enforce it in the cluster-ready navigation guard, redirecting non-admins (no system-admin / read-only admin role) away, mirroring the existing `requiresWorkspaceAdmin` guard.
- Apply the meta to the currently-unguarded, admin-only System pages: clusters, deploy (+detail), auditlogs, registries, addons, addontemplate (catalog), nodeflavor (+detail), fault, preflight (bench), quickstart, user-group.

Note: /nodes is intentionally left open — the nav marks it accessible to all users and the node list is scoped server-side, so guarding it would hide the scoped read-only view from workspace members. This is defense-in-depth for the frontend; server-side authorization remains the primary control.

Relates to #624

### DIFF
--- a/Web/apps/safe/src/router/guards/clusterReady.ts
+++ b/Web/apps/safe/src/router/guards/clusterReady.ts
@@ -45,6 +45,13 @@ export default function setupClusterGuard(router: Router) {
         return next('/')
       }
 
+      // Permission check (system-admin only pages)
+      // These System pages are hidden from the nav for non-admins; also block
+      // direct URL navigation so access control is not merely client-side.
+      if (to.meta?.requiresSystemAdmin && !userStore.hasManagerAccess) {
+        return next('/')
+      }
+
       // Permission check (workspace-admin)
       if (to.meta?.requiresWorkspaceAdmin) {
         if (!wsStore.isCurrentWorkspaceAdmin() && !userStore.hasManagerAccess) {
--- a/Web/apps/safe/src/router/index.ts
+++ b/Web/apps/safe/src/router/index.ts
@@ -56,6 +56,7 @@ const router = createRouter({
           path: '/clusters',
           name: 'Clusters',
           component: () => import('../pages/Clusters/index.vue'),
+          meta: { requiresSystemAdmin: true },
         },
         {
           path: '/cluster/detail',
@@ -195,6 +196,7 @@ const router = createRouter({
           path: '/preflight',
           name: 'Diagnoser',
           component: () => import('../pages/Diagnoser/index.vue'),
+          meta: { requiresSystemAdmin: true },
         },
         {
           path: '/preflight/detail',
@@ -226,17 +228,20 @@ const router = createRouter({
           path: '/fault',
           name: 'Fault',
           component: () => import('../pages/Fault/index.vue'),
+          meta: { requiresSystemAdmin: true },
         },
         {
           path: '/nodeflavor',
           name: 'NodeFlavor',
           component: () => import('../pages/NodeFlavor/index.vue'),
+          meta: { requiresSystemAdmin: true },
         },
         {
           path: '/flavors/:id',
           name: 'nodeFlavorDetail',
           component: () => import('../pages/NodeFlavor/FlavorDetail.vue'),
           props: (route) => ({ id: route.params.id }),
+          meta: { requiresSystemAdmin: true },
         },
         {
           path: '/secrets',
@@ -247,6 +252,7 @@ const router = createRouter({
           path: '/quickstart',
           name: 'QuickStart',
           component: () => import('../pages/QuickStart/index.vue'),
+          meta: { requiresSystemAdmin: true },
         },
         {
           path: '/userquickstart',
@@ -262,6 +268,7 @@ const router = createRouter({
           path: '/registries',
           name: 'ImageReg',
           component: () => import('../pages/ImageReg/index.vue'),
+          meta: { requiresSystemAdmin: true },
         },
         {
           path: '/webshell',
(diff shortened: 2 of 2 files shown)


amd-ama10002-2 (approved):
I think this makes sense. Thanks for the fix!

PR #657, opened 2026-07-09
fix(safe): unblock balance scheduling from unbound pending reservations

chennyiiis:
## Summary

- Fix balance (non-FIFO) scheduling so a workload that is already marked scheduled but whose pods cannot bind to a node (empty admin-node name) no longer holds a full-spec reservation that starves smaller workloads queued behind it. This is what made an Authoring dev box on `project1-dev` sit at `queuePosition: 4` with `In queue - insufficient resources, no cpu available`, even though the workspace uses `queuePolicy: balance`.
- Keep the reservation correct in the short dispatch window: a scheduled workload that has not produced any pod status yet still reserves its full spec, so balance does not over-admit against it across scheduling rounds.
- Only pods already placed on a node keep a reservation; pods still waiting for a node are excluded. FIFO keeps the original full-spec reservation behavior unchanged.

## Root cause

`getLeftTotalResources` treated every already-scheduled, not-yet-running workload as consuming its full spec. Large multi-node jobs (e.g. `utd-multi-node-qqcws-*`) that were dispatched but stuck with pods in `nodeUsage.node == ""` therefore reserved all their CPU/GPU at the SaFE scheduling layer. Authoring dev boxes use `isTolerateAll=true`, which is checked against `leftTotalResources`, so the phantom reservation blocked them with `no cpu available`. The balance queue-skip logic (`continue` vs `break`) was already working; the bug was purely in resource accounting.

## Changes

- `job-manager/pkg/scheduler/scheduler.go`: add `getScheduledPendingUsage`; in balance mode a scheduled-but-not-running workload reserves full spec only while it has no pod status, otherwise it reserves only pods bound to a real node.
- `job-manager/pkg/scheduler/scheduler_resources_test.go`: add coverage for the no-status reservation, mixed bound/unbound, FIFO regression, unbound-release, and an end-to-end `scheduleWorkloads` case where a small tolerate-all workload is scheduled behind an unbound pending reservation.

## Behavior notes

- Balance still respects priority: an unschedulable workload with strictly higher priority than the next one still blocks lower-priority workloads behind it.
- Under balance, multiple large jobs stuck as unbound will optimistically let smaller jobs proceed. This matches the documented balance semantics ("avoid blockage by the front workload in the queue") and the big-job-vs-utilization tradeoff that distinguishes balance from FIFO.

## Test plan

- [x] `go test ./job-manager/pkg/scheduler`
- [x] `go vet ./job-manager/pkg/scheduler`
- [x] `gofmt`
- [ ] Deploy `job-manager` to `project1-dev`, and while an unbound pending large job exists, submit a small Authoring dev box and confirm it is scheduled and reaches Running.

### DIFF
--- a/SaFE/job-manager/pkg/scheduler/scheduler.go
+++ b/SaFE/job-manager/pkg/scheduler/scheduler.go
@@ -542,6 +542,8 @@ func (r *SchedulerReconciler) getLeftTotalResources(ctx context.Context,
 		var err error
 		if w.IsRunning() {
 			totalResource, availableResource, _, err = commonworkload.GetWorkloadResourceUsage(w, filterFunc)
+		} else if !workspace.IsEnableFifo() {
+			totalResource, availableResource, err = getBoundPendingResourceUsage(w, filterFunc)
 		} else {
 			totalResource, err = commonworkload.GetTotalResourceList(w)
 			availableResource = totalResource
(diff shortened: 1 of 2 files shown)


chennyiiis:
Closing per design discussion: an already-scheduled/dispatched workload legitimately holds its reserved resources (下发资源), so a same-priority workload behind it is expected to wait. Making it run ahead is a preemption concern (requires enable-preempt + strictly higher priority), not a balance-queue fix. Releasing those reservations conflicts with the reservation/preemption model and could starve dispatched gang-scheduled jobs.

PR #680, opened 2026-07-16
fix(web): restore workload list filter URL sync on CICD, Authoring, Dynamo

fengshaoyi-amd:
## Summary

The 'persist workload list filters in URL' behavior (originally added in commit e70d49dee) was lost for **CICD**, **Authoring** and **Dynamo** list pages after the pulse-alignment merge (#677). On these pages, clicking the filter/search button ran fetchData directly and never wrote query params to the URL, so:

- filter state was not reflected in the URL, and
- returning from a workload detail page did not restore the previous filters/pagination.

The other six workload list pages (Training, Infer, RayJob, Monarch, TorchFT, SandboxWorkload) still had it via their inline router.replace in onSearch.

## Fix

For CICD, Authoring and Dynamo, mirror the surviving pages' inline pattern:
- **onSearch**: write userName/description/phase/since/until/workloadId/onlyMyself/userId/page/pageSize to the URL via router.replace before fetchData.
- **applyQueryToParams()**: restore searchParams + pagination from route.query.
- restore on mount / workspace change (CICD, Authoring watcher; Dynamo onMounted). Dynamo also gains the missing useRoute import.

## Notes
- The orphaned useWorkloadListQuery composable (unused on main) is left as-is; not needed for this fix.

## Test plan
- [ ] On CICD / Authoring / Dynamo list pages, applying a filter updates the URL query.
- [ ] Navigating into a detail page and back restores the same filters + page.
- [ ] vue-tsc passes; vitest 115/115 pass (verified locally).

### DIFF
--- a/Web/apps/safe/src/pages/Authoring/index.vue
+++ b/Web/apps/safe/src/pages/Authoring/index.vue
@@ -627,17 +627,55 @@ const filterByMyself = () => {
   onSearch({ resetPage: true })
 }
 
+const applyQueryToParams = () => {
+  const q = route.query
+  searchParams.userName = (q.userName as string) || ''
+  searchParams.description = (q.description as string) || ''
+  searchParams.workloadId = (q.workloadId as string) || ''
+  searchParams.onlyMyself = (q.onlyMyself as string) || 'All'
+  searchParams.userId = (q.userId as string) || ''
+  const phaseStr = (q.phase as string) || ''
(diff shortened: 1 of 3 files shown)


fengshaoyi-amd:
Superseded: restoring the complete composable-based URL-query solution (from fix/web-issues-report: e70d49dee/c673a9027/04e127397) instead of the inline rewrite. New PR to follow.

PR #701, opened 2026-07-23
feat(slurm): per-workspace Slurm cluster creation in Primus-SaFE via Slinky

amd-ama10002-2:
## Summary
Adds support for running **Slurm clusters inside a workspace**, backed by the [Slinky](https://github.com/SlinkyProject) `slurm` Helm chart (v1.2.0) while trying to follow the same pattern as Ray Clusters. 

Users with a `Slurm`-scoped workspace can create, view, manage the lifecycle of, and delete a Slurm cluster from the Primus-SaFE console, with SSH access to a login node and optional Slurm accounting.

I have a video recorded demo and a live demo. Please @ me if you are reviewing and want to see either of these. 

## Design overview
A "Slurm cluster" is a **per-workspace Helm release** of the Slinky `slurm` chart, deployed as an `Addon` CR into the workspace namespace. A single shared **slurm-operator** (Slinky, v1.2.0) reconciles the resulting NodeSet/Controller/LoginSet/RestApi CRs.

```mermaid
flowchart LR
    UI["Console: Slurm pages"] --> API["apiserver SlurmCluster handler"]
    API --> Addon["Addon CR (per workspace ns)"]
    Addon --> AC["AddonController: helm install/upgrade/uninstall"]
    AC --> Chart["Slinky slurm chart v1.2.0"]
    Chart --> Pods["slurmctld, restapi, login, slurmd workers, (accounting)"]
    Op["slurm-operator (ns slinky, shared)"] --> Pods
```

- **Backend** (`SaFE/apiserver/pkg/handlers/resources/slurmcluster.go`, `slurmcluster_mariadb.go`, `view/slurmcluster_view.go`) renders the node-pool form into chart values, manages the Addon lifecycle, and enriches list/detail responses with live pod state read from the data cluster.
- **New `Slurm` workspace scope** (`SaFE/apis/pkg/apis/amd/v1/workspace_types.go` + CRD enum) gates the feature.
- **Slurm worker resource accounting** (`SaFE/apiserver/pkg/handlers/resources/workspace.go`): because Slurm pods are not SaFE Workloads, `getSlurmPodUsage` counts `slurmd` worker pods toward the workspace's used quota.
- **Frontend** (`Web/apps/safe/src/pages/SlurmCluster/`): list, detail (stat cards + live pod table), create/clone dialog, and SSH dialog.

## What's implemented
- Create a Slurm cluster from a node-pool form (partitions, node counts, CPU/GPU/memory per node).
- First partition automatically marked **default** (jobs run without `-p`).
- **Detail view** with compute-resource summary and live pod status (controller, login, restapi, workers, accounting).
- **Lifecycle controls**: Stop, Resume, Clone (see "Lifecycle behavior" below).
- **Delete** with full cleanup (see "Lifecycle behavior" below).
- **Optional accounting**: provisions a release-scoped MariaDB + slurmdbd wiring (the chart supplies neither).
- **SSH access** to the login node (`SaFE/apiserver/pkg/handlers/ssh-handlers/k8s.go`).
- **Workspace-volume mounting**: the workspace's configured volumes are mounted into the login and all worker pods.
- **Cluster-name length validation** (client + backend) for the Slurm 40-char internal-name limit.
- Optional per-cluster **image-tag override** applied to slurmctld/slurmd/slurmrestd.
- **Docs**: `docs-site/docs/tasks/create-slurm-cluster.md` task/UAT page (registered in the docs catalog).

## Lifecycle behavior (for reviewers)

**Stop** is a deliberate two-phase operation to avoid stranding GPUs. The Slinky operator drains a worker NodeSet by first creating a maintenance reservation through the REST API (`slurmrestd`), and only then deleting the worker pods.
- Phase 1: scale the worker NodeSets and the login node to zero, while keeping the controller (`slurmctld`) and `slurmrestd` running so the operator can cleanly drain the workers. Tearing `slurmrestd` down at the same time would deadlock the drain and leave workers (and their GPUs) running.
- Phase 2: once the workers have actually drained, scale `slurmrestd` to zero as well.
- The accounting MariaDB is scaled to zero on stop.
- The Addon, its Helm config, and the controller's state PVC are retained, so the cluster stays in the list as history.

**Resume** re-renders the release with the saved replica counts (restoring workers, login, restapi) and scales MariaDB back to one.

**Delete** removes the Addon CR, which triggers a Helm uninstall of the release (removing all chart resources: controller, login, restapi, worker, and accounting pods; the Slinky CRs; services; configmaps; secrets). The handler then explicitly tears down the release-scoped MariaDB (deployment, service, secret, PVC) and deletes the `slurmctld` statesave PVC that Helm uninstall would otherwise orphan (StatefulSet `volumeClaimTemplate` PVCs are not garbage-collected by Kubernetes). Net result: nothing is left behind.

## Testing / UAT
- Created and deleted clusters end-to-end via the console (Playwright): default partition, detail view, stop/resume, clone, name-length validation, and full delete cleanup (verified no leftover PVCs).
- Volume mounting is verified for **HostPath**-type workspace volumes (mounted into the login and all worker pods). **PFS** (ReadWriteMany PVC) volumes are wired through the same code path and are expected to work, but have not been rigorously tested yet.
- See `docs-site/docs/tasks/create-slurm-cluster.md` for the documented UAT steps.

## Where the Slinky image/chart comes from
- Charts: `oci://ghcr.io/slinkyproject/charts/slurm` and `.../slurm-operator`, both **v1.2.0** (`SaFE/charts/primus-safe-cr/templates/addon_template/slurm.1.2.0.yaml`, `slurm-operator.1.2.0.yaml`).
- Container images **default to `ghcr.io/slinkyproject/*`**, pulled directly from ghcr (per the operator template comment), unless a per-cluster image-tag override is set.
- Accounting DB uses `docker.io/library/mariadb:11.4`.
- The `slurm-operator` addon requires **cert-manager** and bundles the Slinky CRDs (ns `slinky`); it is a default catalog addon. The `slurm` chart is a per-workspace chart source, not a default catalog addon.

## Future work / not yet done
- Slinky container images are pulled directly from `ghcr.io` (not mirrored through the local registry proxy); air-gapped installs would need mirroring.
- Worker pods carry pod anti-affinity, so multiple workers won't co-locate on a single node.
- No Go unit tests were added for the new handler (feature was validated via end-to-end UAT).


amd-ama10002-2:
Closing PR. Got some feedback from @weilei0120 . Can reach out to him for deeper discussion, but seems like we need more alignment on what a Slurm feature in Primus-SaFE would look like if we implement something at all.

PR #712, opened 2026-08-13
fix(job-manager): stop charging workspaces for pods that no longer exist

weilei0120:
Three independent bugs in the job-manager syncer, found while investigating why a workload on a development cluster would not start. The scheduler said `In queue - insufficient resources, no cpu available`; nothing was full. Each fix is its own commit.

## Pod records were never released, so they charged the workspace forever

`GetWorkloadResourceUsage` counts every non-terminal pod record, and a record is released only when the delete event for it is handled. A process that ends with events in flight leaves those records at `Running` for good, so the scheduler keeps subtracting pods nobody is running from everyone else's quota. One CI runner set carried hundreds of records for pods that did not exist, which was most of its workspace's CPU quota.

`reconcileVanishedPods` releases them by comparing the records against the pod informer's cache — no API call and no etcd read. It runs at most once per workload per process, which is the whole of what is needed: within a process the delete events are reliable, and the informers replay an add for every live object when the next one starts.

Four preconditions decide whether a record may be released, each there to keep the pass off a pod that is alive:

- **The cache read must have succeeded and the informer must be synced.** An unsynced cache reports empty, so reading "not established" as "none exist" would stop every pod of every workload. It is a distinct return value rather than an empty map.
- **Existence is decided the way the record was attributed.** Pods are matched by the workload id label across all namespaces, and Monarch mesh pods — which carry no label, and whose records are attributed through their mesh object — are resolved the same way the recording path resolves them.
- **The record must be older than a grace period.**
- **The workload must be dispatched and not ended**, and teardown is not evidence that records should go.

There is deliberately no ratio ceiling on one pass: the case that motivated this was nearly every record of one workload, so any such limit would have blocked exactly the repair it was meant to make safe.

## A runner set GitHub never registered held the queue head

Two `AutoscalingRunnerSet` workloads pointing at an organisation whose IP allow list rejects this cluster sat at queue position 0 for a day and a half, each ARC retry failing with the same 403. A `Pending` workload holds the head of its workspace queue, so everything behind it waited on runs that could never start.

The failure leaves no trace on the object — no status, no `runner-scale-set-id`, no event — so the only signal is that the wait has gone on too long. `checkRunnerSetRegistration` fails the workload after ten minutes against a registration that normally takes seconds, which releases the queue and puts the reason on the workload's condition.

## A pod row per finished CI runner

ARC runs one ephemeral pod per CI job and deletes it, and every deletion left a row behind, hydrated from the database on every reconcile. CICD workloads now drop the record on delete, as application workloads already did. This bounds the growth but does not close the leak above, which is what the reconcile is for.

## Also here

`syncGithubAnnotations` is folded into `cluster_client_sets.go`, its only caller. `syncer_status.go` stays where it is: it is the persistence seam three files write through, and the only place that decides whether a status write is offloaded to the database.

## Test plan

- [x] `go build`, `go vet`, `gofmt` clean; `go test ./job-manager/pkg/syncer/ ./job-manager/pkg/utils/ ./common/pkg/workload/` pass
- [x] `job-manager/pkg/server` fails identically before and after these commits (`TestServerStartInited`) and is unrelated
- [x] 20 cases for the reconcile: the release policy in isolation, the cache read against a real `cache.Indexer` (including the cross-namespace listener and Monarch mesh resolution), and the flow — released once per workload, CICD records dropped, undispatched and teardown skipped, ended forgotten, an unknown answer changing nothing and retrying, a stale caller still landing its write
- [x] Two regressions are pinned by tests that fail if the fix is reverted: namespace-scoping the cache read, and deciding mesh pods by label
- [ ] Not verified in a cluster: the reconcile needs a job-manager rollout to observe. The stale records it found were cleaned by hand in the meantime, which restored the workspace's quota

## Not addressed

The IP allow list itself. Until the cluster's egress address is allowed, those runner sets fail on the deadline rather than block the queue — the intended outcome, but not a working CI.


SaFE/job-manager/pkg/syncer/pod_handler.go
```diff
@@ -400,6 +383,217 @@ func (r *SyncerReconciler) removeWorkloadPod(ctx context.Context, message *resou
...
+	// buildObjectLabels -> getRootWorkloadId). Pod records are attached to whichever
+	// workload GetWorkloadId names, which is that same label, so the owner of a
+	// record and this selector are the same value by construction. Selecting on
+	// k8s.object.id instead would release a child workload's records while its pods
+	// run.
+	//
+	// The namespace is not the workspace either: a CICD runner set owns the ARC
+	// listener pod, which runs in the controller's namespace.
+	objs, err := informer.Lister().List(
+		labels.SelectorFromSet(labels.Set{v1.WorkloadIdLabel: adminWorkload.Name}))
```
zoroyihan7, line 558: **The live set is selected by a label that not every pod carries, and Monarch looks like the case that breaks it.**

Existence is decided here from `primus-safe.workload.id` alone, but that is not how pod records are attributed. Monarch takes a different route on both ends:

- `handleResource` deliberately keeps a pod whose workload id label is empty, as long as it carries `monarch.pytorch.org/mesh-name`.
- `getAdminWorkloadAndSyncPod` resolves the owning workload from the mesh object rather than from the pod's own label — which only makes sense if the pod's label is not usable.
- `updateMetadata` returns early for `MonarchMesh`, so the inner `podTemplate` never receives the object labels the dispatcher builds.

If mesh pods indeed carry no `primus-safe.workload.id`, then for a running MonarchJob this selector returns an empty live set, `vanishedPodIds` sees every record as gone, and — since there is deliberately no ratio ceiling — the whole set is released in one pass while the pods are running. That is the "real damage" case the PR guards against for an unsynced cache, reached through a different door: the selector's notion of ownership and the record's notion of ownership are not the same value for every kind.

Could you confirm what labels a Monarch mesh pod actually carries? Two ways out:

1. Skip `IsMonarchJob`/mesh workloads in `reconcileVanishedPods`.
2. Better in my view: stop listing by label and instead confirm each candidate in `gone` individually against the cache (`Lister().ByNamespace(ns).Get(name)`). This runs at most once per workload per process, so even the 254-record case is 254 cache lookups and no API call, and the check is then aligned with the records themselves rather than with a labelling convention that a future change could quietly break.

SaFE/job-manager/pkg/syncer/job_handler.go
```diff
@@ -60,7 +66,16 @@ func (r *SyncerReconciler) handleJob(ctx context.Context,
 		// Errors defined internally are fatal and lead to a terminal state without retry
 		err = jobutils.SetWorkloadFailed(ctx, r.Client, adminWorkload, err.Error())
 	}
-	return result, err
+	if err != nil {
+		return result, err
+	}
+	// Last, so it sees the state this event wrote, and best-effort so it cannot fail
+	// the status sync.
+	if reconcileErr := r.reconcileVanishedPods(ctx, clientSets, adminWorkload, message); reconcileErr != nil {
```
zoroyihan7, line 74: **This does not see the state the event wrote, and in offload mode the expected conflict is not free.**

`handleJobImpl` reassigns its own `adminWorkload` to the deep copy `updateAdminWorkloadByJob` returns, but that reassignment is local to the callee — the pointer here still refers to the pre-update object. So whenever the event actually wrote status, the reconcile runs against a stale copy with a stale resourceVersion, and the comment above ("Last, so it sees the state this event wrote") does not hold.

Two consequences:

- `IsEnd()` is read off the old phase, so a workload this very event marked Failed or Succeeded still gets a full pass.
- More importantly, the design leans on "a conflict is expected and the next event retries", but that is only true for etcd. `patchWorkloadPodStatus` writes the database *before* the resourceVersion-guarded patch, and `writeWorkloadStatusToDB` ends with `DeleteWorkloadPodsNotIn`. When the patch is then rejected, the DB has already been rewritten from the stale pod list while the `NodeUsage` aggregate never lands in etcd — and `GetWorkloadResourceUsage` prefers that aggregate, so the repair silently does not take effect and rows for pods added in the meantime may have been deleted. Later events do heal this, but making the conflict a designed-in path means the window is entered on purpose rather than rarely.

Smallest fix: re-read the workload at the top of `reconcileVanishedPods` via `r.getAdminWorkload`, which also re-hydrates from the DB and costs one extra Get per workload per process. Alternatively, have `handleJobImpl` return the workload it ended with.

SaFE/job-manager/pkg/syncer/job_handler.go
```diff
@@ -27,13 +27,19 @@ import (
...
 
 const (
 	MaxFailedPodsToShow = 3
 	MaxConditionHistory = 30
 )
 
+// runnerSetRegistrationTimeout is how long a dispatched CICD scaling runner set may
+// go unregistered before it is failed. Registration completes in seconds when
+// GitHub is reachable.
+const runnerSetRegistrationTimeout = 10 * time.Minute
```
zoroyihan7, line 41: **A compile-time deadline that ends in a terminal state, with no way to relax it during an incident.**

The IP allow list case should certainly fail, and failing it is what releases the queue. But the signal this acts on — "no `runner-scale-set-id` after 10 minutes" — is also what a cluster-side problem looks like: an ARC controller rollout, an evicted `gha-rs-controller` pod, or an apiserver hiccup that outlasts the window. Because the error is unrecoverable, every dispatched runner set in that window is marked Failed with no retry, and nothing brings them back automatically once ARC recovers.

Would you consider making `runnerSetRegistrationTimeout` (and `vanishedPodGracePeriod` in `pod_handler.go`) configurable through `commonconfig`, as other tunables in the repo are? The defaults can stay as they are; the point is that an operator can widen them while ARC is down instead of watching every runner set get judged.

SaFE/job-manager/pkg/syncer/pod_handler.go
```diff
@@ -400,6 +383,217 @@ func (r *SyncerReconciler) removeWorkloadPod(ctx context.Context, message *resou
...
+		if r.applyPodGone(adminWorkload, id) {
+			changed = true
+		}
+	}
+	if !changed {
+		return nil
+	}
+	// The count is the signal that pod delete events are being lost.
+	klog.Infof("released %d vanished pod record(s) of workload %s, live pods: %d",
+		len(gone), adminWorkload.Name, len(live))
```
zoroyihan7: **Minor: the number logged is the number of candidates, not the number released.**

`applyPodGone` returns false for a record that is already terminal or already Stopped, so `len(gone)` can be larger than what this pass actually changed. Since the comment says the count is the signal that pod delete events are being lost, it is worth counting the releases that really happened — the loop already tracks that in `changed` and could carry a counter instead of a bool.

SaFE/job-manager/pkg/syncer/pod_handler.go
```diff
@@ -400,6 +383,217 @@ func (r *SyncerReconciler) removeWorkloadPod(ctx context.Context, message *resou
...
+	gone := make([]string, 0)
+	for i := range adminWorkload.Status.Pods {
+		p := &adminWorkload.Status.Pods[i]
+		if p.PodId == "" || v1.IsPodTerminated(p) {
+			continue
+		}
+		if _, ok := live[p.PodId]; ok {
+			continue
+		}
+		startedAt, err := timeutil.CvtStrToRFC3339Milli(p.StartTime)
```
zoroyihan7, line 679: **Non-blocking note: the grace period only works because the process happens to run in UTC.**

`StartTime` is produced by `buildWorkloadPodInfo` as `timeutil.FormatRFC3339(pod.Status.StartTime.Time)`. `metav1.Time.UnmarshalJSON` converts to local time, and `FormatRFC3339` writes no zone suffix, so the stored string is local wall-clock while `CvtStrToRFC3339Milli` parses it as UTC. With `TZ=UTC` in the container this is exactly right; anywhere east of UTC the record looks like it started in the future, `startedAt.After(cutoff)` is always true, and nothing is ever released. The direction is the safe one — it never releases a live record — but the whole fix would go quiet without a trace.

The same pattern already exists in `handleRaySubmitterTimeout`, so this is not something this PR introduced and I would not block on it. Flagging it so the assumption is at least written down somewhere.

zoroyihan7:
**The description carries internal environment detail, and this repository is public.**

The code itself is clean on this front — no cluster names, paths, or job ids leak into it. The PR body is a different matter: it names a specific workload and the development cluster it ran on, the internal GitHub organisation behind the allow list, and the manual database cleanup along with the workspace's quota numbers. All of that becomes a permanent part of a public repository once this merges.

The investigation narrative is genuinely the most useful part of this PR and worth keeping. Could you re-tell it in anonymised terms — "a workload on a development cluster", "an organisation whose IP allow list rejects this cluster" — and drop the row counts and the hand-applied SQL? Nothing in the reasoning depends on the specific names.

zoroyihan7:
Reviewed all five commits. Verified locally on `ff78b17`: `go build ./job-manager/...`, `go vet`, `gofmt -l`, and `go test ./job-manager/pkg/syncer/` are all clean, and the four `syncGithubAnnotations` tests moved into `cluster_client_sets_test.go` are a byte-for-byte match with the deleted file, so nothing was lost in the fold.

The diagnosis behind each of the three fixes holds up, and the tests are unusually good — that "not knowing has to mean not acting" is expressed as a distinct return value and covered by its own case is the right instinct for a repair that writes to live state. A few things I liked while reading: `applyPodGone` is a genuine extraction shared with the delete path rather than a copy; re-resolving the index per pod inside the loop correctly handles `applyPodGone` removing entries and `sortWorkloadPods` reordering them; and `TestLivePodNamesSeesPodsInAnyNamespace` pins down a regression that would otherwise be one careless edit away.

I left five separate threads. Two I would like resolved before merge:

- Whether Monarch pods carry `primus-safe.workload.id` at all — if they do not, `livePodNames` returns an empty live set for a running MonarchJob and, with no ratio ceiling, releases every one of its records.
- The workload passed to `reconcileVanishedPods` is a version behind whenever the event wrote status, which makes the "expected conflict" path write the database from a stale pod list before the guarded etcd patch is rejected.

The other three (configurable deadlines, the logged count, the timezone assumption in the grace period) are not blocking.

SaFE/job-manager/pkg/syncer/pod_handler.go
```diff
@@ -400,6 +384,304 @@ func (r *SyncerReconciler) removeWorkloadPod(ctx context.Context, message *resou
...
+
+	// Re-read rather than work from the caller's copy: this event may have written the
+	// status through a deep copy, leaving that copy a version behind. Patching from it
+	// would rewrite the database from a stale pod list -- writeWorkloadStatusToDB ends
+	// with DeleteWorkloadPodsNotIn -- and then have the guarded etcd patch rejected, so
+	// the NodeUsage aggregate the scheduler reads would never land. The Get also
+	// re-hydrates the pod list and gives IsEnd a current phase.
+	adminWorkload, err := r.getAdminWorkload(ctx, adminWorkload.Name)
+	if err != nil || adminWorkload == nil || len(adminWorkload.Status.Pods) == 0 {
+		r.vanishedPodsChecked.Delete(adminWorkload.Name)
```
zoroyihan7: **Blocking: this dereferences a nil workload on every failure path, and the panic takes the process down.**

The short variable declaration on the line above redeclares the parameter rather than shadowing it — `adminWorkload` and the statement are in the same block, so per the spec this is an assignment. `getAdminWorkload` returns `nil` both when the Get fails and when the workload is simply not found (it swallows `IsNotFound` into `nil, nil`). So by the time control reaches `adminWorkload.Name` here, `adminWorkload` is always nil on exactly the paths that lead here.

Reproduced against this commit with a throwaway test — a dispatched workload with one stale record, and a client that does not hold the workload:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20]
syncer.(*SyncerReconciler).reconcileVanishedPods(...)
	SaFE/job-manager/pkg/syncer/pod_handler.go:461
```

This is not contained: `KeyedController.Run` drives the worker through `wait.UntilWithContext`, whose `JitterUntil` defers `runtime.HandleCrash`, and the default handler re-panics — so the goroutine's panic ends the job-manager process. Any transient Get error reaches it without needing a race at all, and the not-found case is reachable for a workload deleted between the event and this re-read.

Keep the name before the re-read:

```go
name := adminWorkload.Name
fresh, err := r.getAdminWorkload(ctx, name)
if err != nil || fresh == nil || len(fresh.Status.Pods) == 0 {
	r.vanishedPodsChecked.Delete(name)
	return err
}
adminWorkload = fresh
```

Worth a test for the workload-not-found path too — the existing cases all re-read a workload that is present, which is why this got through green.

SaFE/job-manager/pkg/syncer/pod_handler.go
```diff
@@ -400,6 +384,304 @@ func (r *SyncerReconciler) removeWorkloadPod(ctx context.Context, message *resou
...
+// the mesh object exactly as getAdminWorkloadAndSyncPod resolves it when recording
+// the pod. Meshes are resolved once each: the lookup is a live read, and a mesh has
+// as many pods as it has nodes.
+func (r *SyncerReconciler) addMeshPodNames(ctx context.Context, clientSets *ClusterClientSets,
+	lister cache.GenericLister, adminWorkload *v1.Workload, live map[string]struct{}) bool {
+	selector, err := labels.Parse(monarchMeshLabel)
+	if err != nil {
+		return false
+	}
+	objs, err := lister.List(selector)
```
zoroyihan7, line 610: **Non-blocking: this resolves every mesh in the cluster, not just this workload's, and one unreadable stranger abandons the pass.**

The selector matches any pod carrying the mesh label anywhere in the cluster, and each distinct mesh is then resolved through `getMonarchMesh`, which is a live `Get` against the data plane (plus a resource-template lookup) rather than a cache read. `TestLivePodNamesResolvesMonarchMeshPods` asserts `meshLookups == 2` and documents this exactly: the unrelated `other-mesh-0` is fetched even though it can never affect the outcome. With N Monarch jobs in a cluster this is N x M live reads, all bunched at process start when the informers replay an add for every object.

The fragility bothers me more than the cost. Because an unresolvable mesh returns false, a mesh being torn down for a completely unrelated job makes *this* workload's answer unknown and burns its pass — and mesh teardown is a normal, frequent event.

Both go away with one observation: a pod that is not in `adminWorkload.Status.Pods` cannot change anything, since `live` is only ever probed for membership by record. So the loop can skip any pod whose name has no record before resolving its mesh, which in the normal case narrows the lookups to the workload's own meshes and makes an unrelated mesh irrelevant rather than fatal.

Separately, `reconcileVanishedPods`'s doc still says existence is read "from the pod informer's cache rather than the API", which is no longer true on this path — worth a word there so the next reader is not surprised by a live Get.

SaFE/job-manager/pkg/syncer/pod_handler.go
zoroyihan7: Thanks — `usable := err == nil && !startedAt.After(now)` closes the direction that mattered: east of UTC the record now falls back to the dispatch timestamp instead of looking eternally young, so the repair no longer goes quiet.

One residual, purely for the record and not something I would hold the PR for: the correction is one-sided. West of UTC the same missing zone makes a record parse as *older* than it is, by the size of the offset, so the five-minute grace period is effectively zero there. That only matters for a pod that is alive yet absent from the live set, which the rest of the design already treats as impossible — so the exposure is small. The real fix is for `StartTime` to carry its zone, which is a wider change than this PR.

SaFE/job-manager/pkg/syncer/job_handler.go
zoroyihan7: This one looks untouched in the new revision — `runnerSetRegistrationTimeout` and `vanishedPodGracePeriod` are still compile-time constants. That is a defensible call and I am not asking to reopen the argument; I would just like to know which way you decided, so the thread can be resolved rather than left hanging.

If the intent is to defer it, a follow-up issue would be enough for me. The concern I would want written down somewhere is the operational one: during an ARC outage longer than the window, every dispatched runner set is marked Failed with no retry and nothing restores them once ARC comes back.

zoroyihan7:
Re-reviewed at `77a5f7a` (the branch was rebased, so I diffed against `ff78b17`, the revision of my first pass — the change is confined to `pod_handler.go` and its tests). Build, `go vet`, `gofmt` and `go test ./job-manager/pkg/syncer/` are all clean again.

Four of the five points are properly addressed, and in a couple of places better than what I suggested:

- **Monarch.** You confirmed the labels are absent and resolved ownership through the mesh object — the same route `getAdminWorkloadAndSyncPod` takes when the record is created — rather than skipping Monarch outright. Deduplicating by mesh and treating an unreadable mesh as unknown are both the right calls, and `TestLivePodNamesResolvesMonarchMeshPods` pins the union of the two lookups.
- **The stale copy.** Re-reading after the `LoadOrStore` is the right place: the memo is claimed first, so the extra Get stays at one per workload per process, and it happens to close a second hole — `reSchedule` clears `Status.Pods`, so a re-scheduled workload now falls out on the fresh read rather than being judged on a list that no longer exists.
- **The logged count** and **the timezone fallback** are both fixed. Folding `p.Phase == Stopped` into `IsPodTerminated`, which already covers it, is a nice bit of tidying on the way past.
- The description reads clean now.

One new blocker, unfortunately, introduced by the re-read itself: on both of its failure paths `adminWorkload` has already been reassigned to nil, so `adminWorkload.Name` panics — and because the worker runs under `wait.UntilWithContext`, the default crash handler re-panics and the job-manager process goes down. A transient Get error is enough to trigger it; no race required. Details and a reproduction are in the thread on that line, along with a three-line fix.

Also left a non-blocking note on `addMeshPodNames` resolving unrelated meshes through live API reads, where one unreadable stranger abandons this workload's pass, and a ping on the deadline-configurability thread just to learn which way you decided.

Happy to re-check as soon as the nil path is fixed — everything else on my list is settled.

SaFE/job-manager/pkg/syncer/job_handler.go
zoroyihan7: Addressed in `77a5f7a` by re-reading the workload inside `reconcileVanishedPods`, placed after the `LoadOrStore` so the extra Get stays at one per workload per process. Resolving this thread.

The nil dereference on that re-read's failure paths is tracked separately in [its own thread](https://github.com/AMD-AGI/Primus-SaFE/pull/712#discussion_r3775241123) so it does not get lost behind a resolved one.

SaFE/job-manager/pkg/syncer/pod_handler.go
weilei0120: Fixed in f456f11e, exactly as suggested — the name is held before the re-read and the fresh copy lands in a separate variable.

Your reading of the redeclaration is right, and it is the part I got wrong: the parameter and that statement share the function body's block, so `:=` assigned to the parameter instead of shadowing it. The not-found path made it a certainty rather than a race, since `getAdminWorkload` swallows `IsNotFound` into `nil, nil`.

Added `TestReconcileVanishedPodsSurvivesAVanishedWorkload` for the path you asked for — a dispatched workload with a stale record against a client that does not hold it. Against the previous commit it reproduces your trace verbatim (`SIGSEGV ... addr=0x20`); it passes now, and also asserts the memo entry is released so a workload that comes back still gets its pass.

SaFE/job-manager/pkg/syncer/pod_handler.go
weilei0120: Taken, in the same commit f456f11e. The loop now skips any pod that no record names before resolving its mesh — your observation that `live` is only ever probed by record is the whole argument, and the fragility you point at is the reason worth acting on: an unrelated job's teardown should not spend this workload's one pass.

In the normal case that leaves the lookups at this workload's own meshes. `TestLivePodNamesResolvesMonarchMeshPods` now asserts `meshLookups == 1` instead of documenting the stranger being fetched, and a new `TestLivePodNamesIgnoresAStrangerMeshItCannotRead` pins the part that changed in kind: a stranger mesh that cannot be read at all leaves the answer usable, with zero lookups.

Also corrected the `reconcileVanishedPods` doc — it now says existence comes from the informer cache except for a mesh workload, whose pods are attributed through a live Get.

SaFE/job-manager/pkg/syncer/job_handler.go
weilei0120: Not doing this in this PR — keeping both as constants for now.

The incident you describe is real, and I am not arguing the deadline cannot misjudge a cluster-side outage. Two reasons to leave it out of this change: the recourse when it does misjudge is a resubmit of a runner set that ARC will register once it is healthy, which is the same action an operator takes after any failed dispatch; and turning these into values read at runtime is a change to the tunable surface — with tests that would then want to override them — that I would rather not fold into a fix whose point is to stop the queue being held. `vanishedPodGracePeriod` in particular only ever delays a release, so widening it is never the lever an incident needs.

Recording it as a follow-up rather than a rejection: if ARC registration proves flaky in practice, the deadline is the first thing that should become configurable, and the default can move on its own. Resolving this thread on that basis.

zoroyihan7:
Re-reviewed at `f456f11`. **Good to merge from my side** — everything I raised across the three rounds is either fixed or settled, and I verified the two fixes independently rather than reading the diff.

**The nil dereference is genuinely gone.** I exercised both branches that reach that exit — a workload deleted between the event and the re-read (`getAdminWorkload` swallowing `IsNotFound` into `nil, nil`) and a re-read that fails outright — and neither panics now: the not-found case returns nil, the failed Get surfaces its error, and both release the memo entry so a workload that comes back still gets its pass. Holding `name` before the re-read and landing the fresh copy in its own variable is exactly the shape that makes the redeclaration harmless, and the comment you added explains *why* the separate variable exists, which is the part a future edit would otherwise undo.

**The mesh scoping change does what it claims.** Filtering by `recorded` before resolving a mesh removes both the cost and the fragility in one move, and `TestLivePodNamesIgnoresAStrangerMeshItCannotRead` pins the behavioural change rather than just the count — a stranger mesh that cannot be read at all now leaves the answer usable, which was the failure mode worth caring about. Correcting the `reconcileVanishedPods` doc to admit the live Get on the mesh path was the right call too.

**On the deadline constants: your reasoning is better than my suggestion, and I withdraw it.** The point that widening `vanishedPodGracePeriod` only ever delays a release — so it is never the lever an incident reaches for — is the part I had not thought through; bundling it with `runnerSetRegistrationTimeout` was sloppy on my part. Deferring the registration deadline with the recourse spelled out and a follow-up recorded is the right resolution for a fix whose purpose is to stop the queue being held.

Verification on this revision: `go build ./job-manager/...`, `go vet`, `gofmt -l` and `go test ./job-manager/pkg/syncer/` all clean. `job-manager/pkg/server` still fails on `TestServerStartInited`, and I checked it on a clean `origin/main` worktree — same failure, same line (`manager.go:151`), so your original assessment that it is unrelated holds.

One limitation for the record, not a request and not something this PR should carry: the repair is event-driven and runs once per workload per process, so it depends on the informers replaying an add at startup. A workload still holding records whose k8s object no longer exists produces no such event and therefore never gets a pass. That gap predates this change — `reSchedule` needs the same event — and the case this PR was built for is covered, but it is worth knowing if stale rows ever turn up again after a rollout.

zoroyihan7 (approved):

PR #713, opened 2026-08-14
fix(resource-manager): keep site directives when syncing nodelocaldns

weilei0120:
## Background

When syncing control-plane addresses, `guaranteeNodeLocalDNS` removed and rebuilt the whole `.:53` server block, dropping the `hosts` entries and any other site directive sharing that block. Data-plane pods rely on those records to resolve internal domains, and the function reconciles every 15 seconds, so a manual restore is reverted.

A second defect: when the Corefile had no `template` marker yet, the previous logic appended a second `.:53` block, which CoreDNS refuses to load.

## Changes

- Rewrite only the `template IN A <system host>` stanza; every other directive in the block is preserved. Duplicate stanzas for the same name are removed rather than left answering with a stale address.
- Reuse the existing root block and create one only when the Corefile has none, keeping that block unique. A root zone is recognised whether written as `.:53` or as a bare `.`, and block headers are matched at brace depth zero so a nested `forward . x {` is not read as a zone list.
- Leave the Corefile untouched when its braces are unbalanced.
- Report a Corefile that already declares the root zone more than once and make no change to it, rather than rewriting the first block and leaving the file unloadable. The warning fires once per transition into that state.
- Read the `bind` address from the Corefile instead of assuming the link-local default, since kubespray exposes it as `nodelocaldns_ip`.
- Add `node_local_dns.forward_to_cluster_dns`. When enabled, the root block forwards to the CoreDNS of the cluster it runs in, over TCP as the `cluster.local` zones do, and the template stanza is removed so CoreDNS owns the system host. The address is read from each cluster's `kube-system/kube-dns` Service, so the setting stays correct across data-plane clusters. Default off, leaving existing sites on their current behaviour.
- Log Corefile rewrites at warning level with a line delta.
- Move the Corefile logic to `nodelocaldns.go`.

## Choosing `forward_to_cluster_dns`

Documented in `values.yaml`, since that is where an operator reads it.

- **CoreDNS becomes a hard dependency.** With the node resolver, pods keep resolving external names while CoreDNS is down and lose only in-cluster resolution. With this enabled, every name the root block does not answer locally goes through CoreDNS, so an outage there takes all resolution down.
- **CoreDNS must resolve the system host before the option is turned on**, because the stanza that answered it locally is removed. CoreDNS orders `template` ahead of `hosts`, so leaving the stanza in place would silently override the record the site maintains.
- **Following the healthy control planes moves to CoreDNS with the stanza.** A static CoreDNS record does not track the probed Endpoints pool, so an apiserver that goes down is no longer dropped automatically. Sites that need that behaviour should keep the option off and put their records in the root block's `hosts`, which this PR now preserves.
- **Confirm CoreDNS resolves against the node** (`dnsPolicy: Default`, the standard deployment). If it forwards back to nodelocaldns the `loop` plugin aborts nodelocaldns at startup and every pod on that node loses DNS.
- **Reverting is not symmetric.** The stanza returns on the next reconcile; the forward directive stays pointed at CoreDNS, because this sync does not overwrite a forward it was not asked to manage.

## Tests

- Corefile functions: sibling directives preserved, block uniqueness, portless root zone reused, nested `forward` line not mistaken for a zone list, duplicate stanzas dropped, stanza removal leaving site directives intact, first write when no template exists, idempotency, FQDN zone header, unbalanced braces, root block counting, `bind` address read from the Corefile, forward rewriting with an existing options sub-block left alone.
- Reconcile level: hosts preservation, duplicate root blocks left untouched, and forwarding to the cluster DNS address with the stanza removed.
- `go build`, `go vet` and `helm lint` pass. `TestConstructLocalDownloadOpsJob`, `TestGuaranteeCICDClusterRoleBindingFull` and `TestGuaranteeForwardIngressFull` fail on the base branch as well and are unrelated to this change.

## Deployment notes

This fix stops further damage; it does not repair a Corefile already damaged by the previous logic.

1. Check for duplicate root blocks first: `kubectl -n kube-system get cm nodelocaldns -o jsonpath='{.data.Corefile}' | grep -cE '^[[:space:]]*\.(:53)? \{'`. A count above 1 means CoreDNS cannot load the file, and restarting will not help. Remove the extra blocks by hand, keeping the one that holds the site's own directives; the sync then takes over on the next reconcile. Until they are removed the sync makes no change and logs a warning naming the cluster.
2. Restore any `hosts` records the previous logic dropped.
3. Restart nodelocaldns to load the result: `kubectl rollout restart ds nodelocaldns -n kube-system`.


SaFE/resource-manager/pkg/resource/cluster_controller.go
```diff
@@ -1092,10 +1092,24 @@ func (r *ClusterReconciler) guaranteeNodeLocalDNS(ctx context.Context, cluster *
...
+	// refuses to load, and which block the site means to keep cannot be inferred, so it is reported
+	// rather than partially rewritten.
+	if blocks := countRootServerBlocks(corefile); blocks > 1 {
+		klog.Warningf("nodelocaldns Corefile in cluster %s declares %d %q blocks, which CoreDNS "+
+			"refuses to load; remove the extra blocks before this sync can manage it",
+			cluster.Name, blocks, rootServerZone)
+		return nil
+	}
+
+	upstream := commonconfig.GetNodeLocalDNSUpstream()
```
zoroyihan7: **One global upstream, but this runs per data-plane cluster**

`GetNodeLocalDNSUpstream()` returns a single platform-wide value, while `guaranteeNodeLocalDNS` runs once per data-plane cluster against that cluster's own client. Each data-plane cluster runs its own CoreDNS, whose ClusterIP depends on that cluster's Service CIDR, so one address cannot be correct for more than one of them.

Once a site sets this option, every other data-plane cluster gets a root block forwarding to an address its nodes cannot reach. The root block handles everything outside `cluster.local`, so that is a total DNS outage on those nodes — a broader failure than the incident this PR fixes.

Suggestion: resolve the address per cluster from that cluster's `kube-system/kube-dns` Service ClusterIP and reduce the config to a boolean. `dnsPolicy: ClusterFirst` pods already reach CoreDNS that way, and #547 established the precedent of resolving the cluster DNS address at runtime rather than configuring it. If a static value must stay, it needs to be per cluster (a Cluster CRD field) rather than one global setting.

zoroyihan7 (commented):
Thanks for the thorough fix. All four defects from the report are addressed, and the extra handling — FQDN zone headers, unbalanced braces, refusing to touch a Corefile with duplicate root blocks, the warning-level rewrite log — goes beyond what was asked. The old helpers are gone and an empty `upstream` keeps existing sites on their current behaviour.

The tests are the strongest part: sibling directives, block uniqueness, idempotency, FQDN zone headers, unbalanced braces, an existing forward sub-block left intact, plus four reconcile-level cases. That is exactly the coverage whose absence let the original defect through. The operational notes are also honest about what the option costs, and putting them in `values.yaml` rather than only in the PR description is the right call.

I left comments inline: two I would like resolved before merge, two worth fixing, and four nits. Everything was reproduced by extracting the new functions verbatim into a standalone program, not read off the diff.

SaFE/resource-manager/pkg/resource/cluster_controller.go
```diff
@@ -1146,56 +1193,211 @@ func (r *ClusterReconciler) getControlPlaneIPs(ctx context.Context) ([]string, e
...
+	for offset := 0; offset < len(corefile); offset = advancePastNewline(corefile, offset) {
+		if isRootZoneHeader(corefile[offset:lineEndIndex(corefile, offset)]) {
+			blocks++
+		}
+	}
+	return blocks
+}
+
+// isRootZoneHeader reports whether line opens a server block whose zone list contains the root zone.
+func isRootZoneHeader(line string) bool {
```
zoroyihan7: **A root zone written without a port defeats the uniqueness guarantee**

This only accepts the `.:53` token, but CoreDNS also accepts a bare `.` zone, which defaults to port 53. For such a Corefile `countRootServerBlocks` returns 0 and `findRootServerBlock` finds nothing, so `upsertTemplateStanza` falls through to appending a new block:

```go
corefile := "cluster.local:53 {\n    kubernetes\n}\n" +
	". {\n    hosts {\n        10.9.9.1 intra-a.example.com\n        fallthrough\n    }\n" +
	"    forward . /etc/resolv.conf\n}\n"

out := upsertTemplateStanza(corefile, "foo.local", []string{"10.0.0.1"}, defaultDNSUpstream, false)
// out declares both ". {" and ".:53 {": two root zones on the same bind address, which is
// exactly the Corefile CoreDNS refuses to load that this PR sets out to prevent.
```

kubespray does not write it this way, so the likelihood is low, but block uniqueness is the foundation the whole fix rests on. Accepting `.` alongside `.:53` here (and therefore in `countRootServerBlocks`) closes the hole.

SaFE/resource-manager/pkg/resource/cluster_controller.go
```diff
@@ -1092,10 +1092,24 @@ func (r *ClusterReconciler) guaranteeNodeLocalDNS(ctx context.Context, cluster *
...
+		klog.Warningf("nodelocaldns Corefile in cluster %s declares %d %q blocks, which CoreDNS "+
+			"refuses to load; remove the extra blocks before this sync can manage it",
+			cluster.Name, blocks, rootServerZone)
+		return nil
+	}
+
+	upstream := commonconfig.GetNodeLocalDNSUpstream()
+	// force_tcp only makes sense together with a configured upstream.
+	forceTCP := upstream != "" && commonconfig.GetNodeLocalDNSForceTCP()
+	desired := upsertTemplateStanza(corefile, dnsName, controlPlaneIPs, dnsUpstreamOrDefault(upstream), forceTCP)
```
zoroyihan7: **The template stanza is still force-injected, with no way to opt out**

The sync no longer deletes what it does not own, which was the point of the fix. But it still writes its own `template IN A <system host>` stanza into the root block every reconcile, and there is no setting to turn that off. A site that has deliberately moved all name policy into CoreDNS — which is precisely what `upstream` enables — gets the one piece of hardcoding it was trying to remove reinstated every 15 seconds.

This is not only style: CoreDNS orders `template` ahead of `hosts` in its plugin chain, so the injected stanza answers the system host locally and wins over any record the site maintains for that name in CoreDNS. Such a record silently has no effect.

Suggestion: skip the injection when `upstream` is set. Once the root block forwards to CoreDNS the system host resolves there anyway, so the local stanza is redundant. A separate `manage_template` flag would also work, but tying it to `upstream` keeps the option count down and lets the two settings describe one coherent choice.

SaFE/resource-manager/pkg/resource/cluster_controller.go
```diff
@@ -1146,56 +1193,211 @@ func (r *ClusterReconciler) getControlPlaneIPs(ctx context.Context) ([]string, e
...
+		if zone == rootServerZone {
+			return true
+		}
+	}
+	return false
+}
+
+// findTemplateStanza returns the byte range of the "template IN A <dnsName>" stanza, spanning whole
+// lines so removing it leaves no partial line behind.
+func findTemplateStanza(corefile, dnsName string) (int, int, bool) {
```
zoroyihan7: **Only the first template stanza is updated (regression against the previous logic)**

This returns on the first match, whereas `stripDNSServerBlock` looped until no marker remained. With two stanzas for the same name in one root block, the first is retargeted and the second keeps its stale addresses:

```go
corefile := ".:53 {\n" +
	"    template IN A foo.local {\n        answer \"{{ .Name }} 60 IN A 10.9.9.9\"\n        fallthrough\n    }\n" +
	"    template IN A foo.local {\n        answer \"{{ .Name }} 60 IN A 10.9.9.9\"\n        fallthrough\n    }\n" +
	"    forward . /etc/resolv.conf\n}\n"

out := upsertTemplateStanza(corefile, "foo.local", []string{"10.0.0.1"}, defaultDNSUpstream, false)
// the second stanza still answers 10.9.9.9
```

A healthy sync does not produce such a Corefile, but repairing what the previous logic left behind is part of this PR's job, so extra stanzas are worth removing rather than leaving to answer with a dead address.

SaFE/resource-manager/pkg/resource/cluster_controller.go
```diff
@@ -1146,56 +1193,211 @@ func (r *ClusterReconciler) getControlPlaneIPs(ctx context.Context) ([]string, e
...
+		dnsServerBlockPrefix, renderTemplateStanza(dnsName, ips),
+		dnsStanzaIndent, dnsStanzaIndent, dnsStanzaIndent, dnsStanzaIndent,
+		dnsStanzaIndent, nodeLocalDNSBindIP,
+		renderForwardDirective(upstream, forceTCP), dnsStanzaIndent)
+}
+
+// upsertTemplateStanza points the template stanza for dnsName at the given addresses. Only that
+// stanza is rewritten, so hosts entries and any other directive sharing the block survive. A root
+// block is created only when the Corefile has none, keeping that block unique.
+func upsertTemplateStanza(corefile, dnsName string, ips []string, upstream string, forceTCP bool) string {
```
zoroyihan7: **`upstream` and `forceTCP` are redundant parameters here**

Both are read only on the branch that creates a root block from scratch, and `applyForwardUpstream` rewrites the same directive immediately afterwards. Creating the block with `defaultDNSUpstream` produces a byte-identical result:

```go
noRoot := "cluster.local:53 {\n    kubernetes\n}\n"
a := applyForwardUpstream(upsertTemplateStanza(noRoot, "foo.local", ips, "192.168.0.10", true), "192.168.0.10", true)
b := applyForwardUpstream(upsertTemplateStanza(noRoot, "foo.local", ips, defaultDNSUpstream, false), "192.168.0.10", true)
// a == b
```

Dropping them takes the signature from five parameters to three and leaves each function with one job: `upsertTemplateStanza` owns the stanza, `applyForwardUpstream` owns the forward directive.

SaFE/resource-manager/pkg/resource/cluster_controller.go
```diff
@@ -1211,6 +1413,31 @@ func dnsServerBlockEnd(corefile string, start int) int {
 	return -1
 }
 
+// lineStartIndex returns the index of the first byte on the line containing pos.
+func lineStartIndex(s string, pos int) int {
```
zoroyihan7: **Consider moving the Corefile logic into its own file**

`cluster_controller.go` is already past 1400 lines and this adds roughly 270 more. Corefile parsing has nothing to do with the cluster lifecycle, and `lineStartIndex`, `lineEndIndex`, `advancePastNewline` and `summarizeCorefileChange` are general-purpose text helpers rather than controller code.

A `nodelocaldns.go` in the same package, with the text helpers in utils, would be a pure move with no behaviour change.

SaFE/resource-manager/pkg/resource/cluster_controller.go
```diff
@@ -1092,10 +1092,24 @@ func (r *ClusterReconciler) guaranteeNodeLocalDNS(ctx context.Context, cluster *
...
-	base := strings.TrimRight(stripDNSServerBlock(corefile, dnsName), " \t\n")
-	desired := base + "\n" + buildDNSServerBlock(dnsName, controlPlaneIPs)
+	// The recorded addresses must follow the healthy control planes, otherwise data-plane nodes
+	// keep resolving the system host to an apiserver that is gone. Only the template stanza is
+	// rewritten so site-specific directives sharing the block are preserved.
+	// Earlier versions of this sync could append a second root block. Such a Corefile is one CoreDNS
+	// refuses to load, and which block the site means to keep cannot be inferred, so it is reported
+	// rather than partially rewritten.
+	if blocks := countRootServerBlocks(corefile); blocks > 1 {
+		klog.Warningf("nodelocaldns Corefile in cluster %s declares %d %q blocks, which CoreDNS "+
```
zoroyihan7: **This warning is not throttled**

A damaged Corefile logs this every 15 seconds per cluster for as long as the damage lasts. The condition only changes through manual repair, so logging it once per transition — or at a lower verbosity after the first occurrence — keeps it visible without flooding the log.

SaFE/resource-manager/pkg/resource/cluster_controller.go
```diff
@@ -1146,56 +1193,211 @@ func (r *ClusterReconciler) getControlPlaneIPs(ctx context.Context) ([]string, e
...
+const (
+	// rootServerZone is the zone list opening the nodelocaldns server block that answers every
+	// query no more specific zone claims.
+	rootServerZone = ".:53"
+	// dnsServerBlockPrefix opens the CoreDNS server block serving rootServerZone.
+	dnsServerBlockPrefix = rootServerZone + " {"
+	// dnsStanzaIndent is the indentation CoreDNS Corefiles use for directives inside a block.
+	dnsStanzaIndent = "    "
+	// nodeLocalDNSBindIP is the link-local address nodelocaldns listens on.
+	nodeLocalDNSBindIP = "169.254.25.10"
```
zoroyihan7: **`nodeLocalDNSBindIP` is hardcoded**

kubespray exposes this as `nodelocaldns_ip`, so a site that changed it would get a block bound to the wrong address. Only the rare create-from-scratch path is affected; reading it from the existing block's `bind` directive when one is present would cover it.

SaFE/resource-manager/pkg/resource/cluster_controller.go
weilei0120: Fixed in 9fe5c49. You are right that one address cannot be correct for more than one data-plane cluster, and the failure mode is worse than the incident this PR fixes.

Took the runtime-resolution route rather than a CRD field: `node_local_dns.upstream` and `node_local_dns.force_tcp` are gone, replaced by a single boolean `node_local_dns.forward_to_cluster_dns`. When it is on, `getClusterDNSIP` reads the ClusterIP from that cluster's own `kube-system/kube-dns` Service and errors out if it is absent or headless. `force_tcp` is no longer a knob either: the target is always a cluster Service now, so TCP is unconditional, matching the `cluster.local` zones.

SaFE/resource-manager/pkg/resource/cluster_controller.go
weilei0120: Fixed in 9fe5c49. The plugin-order point is the decisive one — a site record that silently has no effect is worse than an absent feature.

Tied it to the switch as suggested rather than adding `manage_template`. With `forward_to_cluster_dns` on, `desiredNodeLocalDNSCorefile` calls `removeTemplateStanza` instead of upserting, so an already-injected stanza is also cleared. Leaving a stale stanza would have been the worse of the two states, since it keeps answering with addresses the sync no longer considers healthy while outranking CoreDNS.

Documented in `values.yaml` that CoreDNS must then resolve the system host itself, since that responsibility moves with the switch.

SaFE/resource-manager/pkg/resource/cluster_controller.go
weilei0120: Fixed in 9fe5c49. Reproduced your case before fixing it.

`rootZoneTokens` now accepts both `.` and `.:53`. That alone would have introduced a false positive: `forward . /etc/resolv.conf {` carries the same tokens as a root zone list. So `findRootServerBlock` and `countRootServerBlocks` now track brace depth and only consider lines at depth zero, which is the real discriminator between a zone list and a nested directive. `braceDelta` relies on CoreDNS template actions such as `{{ .Name }}` being balanced.

Covered by `TestUpsertTemplateStanzaReusesPortlessRootBlock` plus two `countRootServerBlocks` cases, including the nested `forward` line.

SaFE/resource-manager/pkg/resource/cluster_controller.go
weilei0120: Fixed in 9fe5c49. Agreed that repairing what the previous logic left behind is in scope.

`upsertTemplateStanza` now retargets the first stanza and then calls `stripTemplateStanzas` from just past it, so duplicates are removed rather than left answering with a dead address. `TestUpsertTemplateStanzaDropsDuplicateStanzas` asserts one stanza remains, the stale address is gone and the braces stay balanced.

SaFE/resource-manager/pkg/resource/cluster_controller.go
weilei0120: Fixed in 9fe5c49. Signature is now `upsertTemplateStanza(corefile, dnsName, ips)`. A created block starts on `defaultDNSUpstream` and `applyForwardUpstream` owns that line, so each function has one job as you described.

SaFE/resource-manager/pkg/resource/cluster_controller.go
weilei0120: Fixed in 9fe5c49. Moved to `nodelocaldns.go` in the same package; `cluster_controller.go` is back to about 1170 lines. Kept the text helpers next to their only caller rather than in utils, since they are a Corefile-parsing detail — happy to move them if you prefer utils.

SaFE/resource-manager/pkg/resource/cluster_controller.go
weilei0120: Fixed in 9fe5c49. A `damagedNodeLocalDNS sync.Map` on the reconciler makes the warning fire once per transition into the damaged state, dropping to V(4) while it persists, and the entry is cleared when the Corefile parses again so a relapse is reported afresh.

SaFE/resource-manager/pkg/resource/cluster_controller.go
weilei0120: Fixed in 9fe5c49. `nodeLocalDNSBindIP` reads the first `bind` directive in the Corefile and only falls back to the link-local default when none is present. Any block will do since nodelocaldns binds them all to the same address. Covered by `TestNodeLocalDNSBindIP` and `TestRenderRootServerBlockUsesCorefileBindIP`.

SaFE/resource-manager/pkg/resource/nodelocaldns.go
```diff
@@ -0,0 +1,336 @@
...
+	return advancePastNewline(corefile, end), true
+}
+
+// hasUndelimitedBlock reports whether the Corefile declares a root zone or a template stanza for
+// dnsName whose bounds could not be resolved, which means its braces are unbalanced.
+func hasUndelimitedBlock(corefile, dnsName string) bool {
+	if strings.Contains(corefile, templateStanzaMarker+dnsName) {
+		return true
+	}
+	return countRootServerBlocks(corefile) > 0
```
zoroyihan7: **An unbalanced line hides the root block, and the fallback appends a second one**

Depth tracking removes the `forward . x {` false positive, but it introduces a new way to miss the root block: any line whose braces do not balance shifts `depth` for the rest of the file, and every later root zone header is then skipped. `hasUndelimitedBlock` derives from the same depth-tracked count, so it does not catch that either, and `upsertTemplateStanza` falls through to appending a second root block — the exact Corefile this PR exists to prevent.

A partially commented-out block is the realistic way to end up with an unbalanced line:

```go
corefile := "# TODO: tidy this {\n" +
	"cluster.local:53 {\n    kubernetes\n}\n" +
	".:53 {\n    hosts {\n        10.9.9.1 intra-a.example.com\n        fallthrough\n    }\n" +
	"    forward . /etc/resolv.conf\n}\n"

countRootServerBlocks(corefile)  // 0, though the file declares one
upsertTemplateStanza(corefile, "foo.local", []string{"10.0.0.1"})
// the result declares two ".:53 {" blocks
```

This is worse than the portless case it replaced, because the guard in `guaranteeNodeLocalDNS` stays blind afterwards: `countRootServerBlocks` still returns 0 for the damaged result, so the sync neither reports it nor backs off, and it keeps managing a file CoreDNS will not load.

Suggestion: keep the depth-tracked scan for locating the block, but make `hasUndelimitedBlock` conservative again — a line-based scan that ignores depth, or a global brace-balance check. Refusing to append whenever a root zone token appears anywhere in the file is enough, and it restores the "leave a Corefile we do not fully understand alone" rule the rest of the function follows.

zoroyihan7 (commented):
Re-reviewed at `3268346`. All eight comments from the previous round are addressed. I checked each one by running the new `nodelocaldns.go` rather than reading the diff: a portless root zone now reuses the existing block, a nested `forward . x {` is not mistaken for a zone list, duplicate stanzas are dropped with the braces left balanced, `bind` is read from the Corefile with the link-local default as fallback, a created block leaves its forward line to `applyForwardUpstream`, and with the switch on the stanza is removed, the forward retargets with `force_tcp`, site `hosts` survive and the result is idempotent.

Replacing `upstream` and `force_tcp` with the single `forward_to_cluster_dns` boolean is a better shape than what I suggested. Reading the address from each cluster's own Service removes the whole class of misconfiguration rather than documenting around it, and folding `force_tcp` in is right once the target is always a cluster Service. The `values.yaml` section on what the switch gives up — in particular that following the healthy control planes moves to CoreDNS with the stanza — is the kind of thing that prevents the next incident.

Three new findings inline, one of which I would like resolved before merge: the new depth tracking has a failure mode that ends in the same duplicate root block this PR exists to prevent. The other two are small.

Unrelated: the Copilot comment about `default` being called with one argument was already fixed in `18a62e7f`, where the template moved to `default dict`. It can be dismissed.

SaFE/resource-manager/pkg/resource/cluster_controller.go
```diff
@@ -1092,24 +1093,96 @@ func (r *ClusterReconciler) guaranteeNodeLocalDNS(ctx context.Context, cluster *
...
+	// Warning level: this ConfigMap is also edited by hand, so every rewrite must be traceable.
+	klog.Warningf("updated nodelocaldns Corefile in cluster %s (%s)",
+		cluster.Name, summarizeCorefileChange(corefile, desired))
 	return nil
 }
 
+// desiredNodeLocalDNSCorefile computes the Corefile this sync wants for the given cluster.
+func (r *ClusterReconciler) desiredNodeLocalDNSCorefile(ctx context.Context, cluster *v1.Cluster,
+	corefile, dnsName string) (string, error) {
+	if !commonconfig.IsNodeLocalDNSForwardToClusterDNS() {
```
zoroyihan7: **Turning the switch off does not restore resolution**

`values.yaml` is upfront that disabling this leaves the forward pointed at CoreDNS, and the reasoning — do not overwrite a forward the sync was not asked to manage — is sound in general. What makes it awkward here is that the situation which prompts someone to turn the switch off is usually a CoreDNS outage, and in that situation the revert does not actually restore resolution: the stanza comes back so the system host resolves again, but every other name still goes to a CoreDNS that is down. Since nodelocaldns does not hot-reload, the operator then has to hand-edit and roll the DaemonSet as well, which is a slow path during an incident.

The sync can recognise its own writing here without keeping any state: if the switch is off and the root block's `forward .` points at that cluster's kube-dns ClusterIP, it wrote that line and can restore `/etc/resolv.conf`. That makes the option symmetric, which is worth having for a switch whose documented downside is losing all resolution.

SaFE/resource-manager/pkg/resource/nodelocaldns.go
```diff
@@ -0,0 +1,336 @@
...
+	bodyStart := advancePastNewline(corefile, blockStart)
+	closingLine := lineStartIndex(corefile, blockEnd-1)
+	for offset := bodyStart; offset < closingLine; {
+		lineEnd := lineEndIndex(corefile, offset)
+		fields := strings.Fields(corefile[offset:lineEnd])
+		if len(fields) >= 2 && fields[0] == "forward" && fields[1] == "." {
+			// A trailing brace opens a sub-block the site owns, so only the address is retargeted
+			// and whatever options it already carries stay in place.
+			if fields[len(fields)-1] == "{" {
+				return corefile[:offset] + dnsStanzaIndent + "forward . " + upstream + " {" + corefile[lineEnd:]
```
zoroyihan7: **force_tcp is skipped when the site's forward already has a sub-block**

Now that `force_tcp` is unconditional rather than a knob, this branch is the one case where the guarantee does not hold: a site whose forward already carries options gets its address retargeted to a cluster Service but keeps a UDP path.

```go
corefile := ".:53 {\n    forward . /etc/resolv.conf {\n        max_concurrent 1000\n    }\n}\n"
applyForwardUpstream(corefile, "192.168.0.10", true)
// forward . 192.168.0.10 { max_concurrent 1000 }  -- no force_tcp
```

The previous revision documented this caveat under the `force_tcp` value; that sentence went away with the option, so the behaviour is now undocumented as well. Inserting `force_tcp` into the existing sub-block when it is absent would keep the site's other options and make the guarantee hold in every case; short of that, the caveat is worth a line in `values.yaml`.

SaFE/resource-manager/pkg/resource/cluster_controller.go
```diff
@@ -1092,24 +1093,96 @@ func (r *ClusterReconciler) guaranteeNodeLocalDNS(ctx context.Context, cluster *
...
+	}
+	// CoreDNS orders template ahead of hosts, so a stanza left behind would override the records
+	// the site maintains in CoreDNS for the same name.
+	desired := removeTemplateStanza(corefile, dnsName)
+	// force_tcp matches how the cluster.local zones reach CoreDNS.
+	return applyForwardUpstream(desired, clusterDNSIP, true), nil
+}
+
+// getClusterDNSIP returns the CoreDNS Service address inside the given data-plane cluster.
+func (r *ClusterReconciler) getClusterDNSIP(ctx context.Context, cluster *v1.Cluster) (string, error) {
```
zoroyihan7: **The client factory is resolved twice**

`guaranteeNodeLocalDNS` has already resolved the factory and holds a `clientSet` for this cluster by the time it gets here. Passing that in removes the second lookup and the second error path for a failure the caller has already ruled out.

SaFE/resource-manager/pkg/resource/cluster_controller.go
```diff
@@ -1092,24 +1093,96 @@ func (r *ClusterReconciler) guaranteeNodeLocalDNS(ctx context.Context, cluster *
...
 
 	cm.Data["Corefile"] = desired
-	if _, err = clientSet.CoreV1().ConfigMaps("kube-system").Update(ctx, cm, metav1.UpdateOptions{}); err != nil {
+	if _, err = clientSet.CoreV1().ConfigMaps(nodeLocalDNSNamespace).Update(
+		ctx, cm, metav1.UpdateOptions{}); err != nil {
 		return fmt.Errorf("failed to update nodelocaldns ConfigMap in cluster %s: %w", cluster.Name, err)
 	}
-	klog.Infof("updated nodelocaldns Corefile in cluster %s with %s -> %v", cluster.Name, dnsName, controlPlaneIPs)
+	// Warning level: this ConfigMap is also edited by hand, so every rewrite must be traceable.
+	klog.Warningf("updated nodelocaldns Corefile in cluster %s (%s)",
```
zoroyihan7: **The rewrite log no longer names the addresses**

This used to name the addresses it wrote and now reports only a line delta. `+2/-2 lines` says something changed but not what the system host resolves to afterwards, which is the fact worth having when auditing this ConfigMap after an incident. Keeping the delta and adding the control-plane addresses on the upsert path, or the CoreDNS address on the other, would cover both.

SaFE/resource-manager/pkg/resource/nodelocaldns.go
weilei0120: Fixed in 8dbd927. This one was mine and you are right that it was worse than what it replaced — thanks for chasing the depth tracking through to its consequence.

The root cause was that `braceDelta` counted braces CoreDNS never sees. Brace counting now stops at an unquoted `#`, via a `codeSpan` helper used by `braceDelta` and `isRootZoneHeader`; `matchingBraceEnd` got the same treatment as a byte-level state machine, since a comment inside a block would have thrown off the block end too. Your example now reports one root block and the upsert joins it.

Took the conservative suggestion as well, because comment handling alone would only cover the cases I thought of: `hasUndelimitedBlock` is back to a depth-agnostic line scan, so any root zone token at the start of a line blocks the append path outright.

Also closed the "stays blind afterwards" half. `unmanageableCorefileReason` now refuses on unbalanced braces in addition to duplicate root blocks, so a file this sync cannot parse is reported and left alone rather than managed. Covered by `TestCommentedBraceDoesNotHideRootBlock`, `TestUnbalancedCorefileIsNeverAppendedTo` and `TestUnmanageableCorefileReason`.

SaFE/resource-manager/pkg/resource/nodelocaldns.go
weilei0120: Fixed in 8dbd927. Agreed — an unconditional guarantee with one silent exception is worse than the knob it replaced.

`applyForwardUpstream` now parses the existing sub-block into its options, keeps them, and appends `force_tcp` when absent. Your `max_concurrent 1000` case keeps that option and gains `force_tcp`, and reapplying does not add it twice. `TestApplyForwardUpstreamAddsForceTCPToExistingSubBlock` asserts both.

SaFE/resource-manager/pkg/resource/cluster_controller.go
weilei0120: Fixed in 8dbd927. The point that a revert happens during a CoreDNS outage, and that the old revert did not restore resolution in exactly that situation, is what settled it.

Implemented the attribution you described: with the switch off, `revertClusterDNSForward` reads the root block's forward target and only rewrites it when it equals this cluster's kube-dns ClusterIP. A forward aimed anywhere else is left alone, and the Service is only looked up when the target is not already `/etc/resolv.conf`, so the default path adds no API call.

`revertRootForward` also drops a sub-block holding nothing but `force_tcp`, since that is this sync's own writing, while keeping a sub-block that carries anything else. `values.yaml` now describes the revert as symmetric, with the DaemonSet roll still required. Covered by `TestGuaranteeNodeLocalDNSRevertRestoresNodeResolver` and `TestGuaranteeNodeLocalDNSKeepsForwardThisSyncDidNotWrite`.

SaFE/resource-manager/pkg/resource/cluster_controller.go
weilei0120: Fixed in 8dbd927. `getClusterDNSIP` now takes the `kubernetes.Interface` the caller already holds, so the factory is resolved once per reconcile and that error path is gone.

SaFE/resource-manager/pkg/resource/cluster_controller.go
weilei0120: Fixed in 8dbd927. `desiredNodeLocalDNSCorefile` returns a description of the target alongside the Corefile, so the log reads `safe.local now resolves via control planes [10.0.0.1 10.0.0.2] (+2/-2 lines)`, or `via CoreDNS at 192.168.0.10` on the other path. The skip path at V(4) names it too.

SaFE/resource-manager/pkg/resource/cluster_controller.go
weilei0120: Correcting my earlier reply: the `kube-system/kube-dns` lookup I described did not survive contact with a real cluster, and the final shape in 4bdf6ee is different.

On the site this started from, kubespray names that Service `coredns`, not `kube-dns`, so the lookup found nothing: with the switch on the sync would have errored every reconcile and never written anything. Hardcoding `coredns` instead would only have moved the assumption.

`clusterDNSAddress` now reads the address out of the Corefile being reconciled, from the `forward .` of the zones nodelocaldns already aims at CoreDNS. Those zones are configured at install time with that cluster's own cluster DNS, so the value is per cluster by construction, with no Service name assumption, no API call and no RBAC. The reverse zones have fixed names, so it still works when the cluster domain is customised. When no such zone exists the sync errors for that cluster alone and leaves its Corefile untouched, so one odd cluster cannot affect the others.

This also removes the second client factory lookup you raised separately, more completely than passing the clientSet through did: there is no Service call left at all.

The only platform-wide value is the boolean itself, which is a policy choice rather than an address. Verified against the live Corefile: the discovered address matches the one the cluster zones use, an unchanged Corefile round-trips byte-identical, and turning the switch off restores the original exactly.

SaFE/resource-manager/pkg/resource/cluster_controller.go
```diff
@@ -1087,29 +1088,102 @@ func (r *ClusterReconciler) guaranteeNodeLocalDNS(ctx context.Context, cluster *
...
+	if err != nil {
+		return "", "", err
+	}
+	upsertTemplateStanza(root, dnsName, controlPlaneIPs)
+	// Reverting must restore resolution, not just the stanza: the usual reason to turn forwarding
+	// off is that CoreDNS is unreachable, and every other name would keep going there.
+	if clusterDNS, ok := clusterDNSAddress(cf); ok {
+		if plugin := findForward(root); plugin != nil && len(plugin.Args) >= 2 && plugin.Args[1] == clusterDNS {
+			revertRootForward(root)
+		}
```
zoroyihan7: **The default path reverts a forward this sync never wrote**

Address equality is not ownership. `clusterDNSAddress` returns whatever the cluster zones forward to, and a site that pointed its root block at its own CoreDNS by hand has precisely that address in the root `forward .`. This branch reads that as its own writing and undoes it — with the switch at its default `false`, i.e. on every deployment that has not opted into anything.

Ran the converged Corefile from the remediation this PR came out of (root block aimed at CoreDNS, template and hosts removed, name policy moved into CoreDNS) through the sync with the switch off:

```diff
 .:53 {
+    template IN A safe.example.com {
+        answer "{{ .Name }} 60 IN A 10.0.0.1"
+        fallthrough
+    }
     errors
     cache 30
     reload
     loop
     bind 169.254.25.10
-    forward . 192.168.0.3 {
-        force_tcp
-    }
+    forward . /etc/resolv.conf
     prometheus :9253
 }
```

Every pod on those nodes goes back to the node resolver, and the records the site moved into CoreDNS stop being reachable — the failure this PR exists to prevent, now reached through the default configuration. It is a wider blast radius than the original defect, too: that one only rewrote the block it recognised as its own, this one rewrites a directive the site owns outright, and the only signal is a `+4/-3 lines` warning that reads like routine drift.

The principle stated a few lines up in `unmanageableCorefileReason` — which parts the site means to keep cannot be inferred — holds here as well. Record ownership instead of inferring it: set an annotation when this sync points the root at CoreDNS (`safe.amd.com/nodelocaldns-forward: cluster-dns` or similar), revert only when it is present, and clear it as part of reverting. A forward the site wrote is then left alone whatever address it carries, and the revert still fires after a restart of the manager, which a flag transition held in memory would not.

Carries forward the revert thread: restoring resolution rather than just the stanza was the right call, the attribution is what does not hold.

SaFE/resource-manager/pkg/resource/nodelocaldns.go
```diff
@@ -0,0 +1,334 @@
...
+			continue
+		}
+		sb.WriteString(server.ToString())
+	}
+	out := sb.String()
+	if !strings.HasSuffix(original, "\n") {
+		return strings.TrimRight(out, "\n")
+	}
+	return out
+}
```
zoroyihan7, line 79: **The serialiser drops every comment, so the first sync after upgrade deletes them**

The doc says this serialises a Corefile the way it was written. It reprints it from the parsed structure, and the caddyfile lexer discards comments, so they are gone before `renderCorefile` is reached. Round-tripping this:

```
# managed by ops, do not edit without telling #infra
.:53 {
    # split-horizon workaround, see ticket OPS-1234
    hosts {
        10.9.9.1 intra-a.example.com
        fallthrough
    }
    errors
    forward . /etc/resolv.conf
}
```

returns the same file with both comment lines removed. Since `desired == corefile` is the only write guard, a Corefile the round trip cannot reproduce differs from itself and is rewritten on the first reconcile after upgrade, with nothing about the stanza having changed. This ConfigMap is hand-edited by definition — that is the premise of the PR — and the comment explaining why a hosts entry exists is what the next person needs when the next incident lands. `TestRenderCorefileRoundTripsUnchanged` holds because its fixture is comment-free and canonically indented; it does not cover the files this is about.

Same applies to anything the two-level plugin model cannot hold. `forward . x { tls { server_name y } }` comes back as a bare `tls`, the inner block dropped. Braces balance, the parse succeeds, the content is gone.

Suggest making "I cannot reproduce this file" a manageability condition alongside the two already in `unmanageableCorefileReason`, which already parses, so it costs one render and a comparison. Content loss rather than byte equality is the right test — otherwise indentation style alone would take a Corefile out of management:

```go
// A Corefile this serialiser cannot reproduce holds something it would drop: a comment, or a
// block deeper than the plugin options the parser models. Reformatting is acceptable, losing a
// line is not.
if lost := droppedLines(text, renderCorefile(cf, text)); len(lost) > 0 {
    return fmt.Sprintf("uses syntax this sync cannot preserve (%q)", lost[0])
}
```

`droppedLines` is the trimmed-line comparison `summarizeCorefileChange` already performs, returning the lines rather than counting them; the two can share it.

Checked what that refuses in practice: the kubespray Corefile, the converged one from the site, tab and two-space indentation, blank lines between blocks and CRLF all round-trip clean and stay managed. Only comments and deeper nesting are refused. Refusing is loud and recoverable — the operator sees the warning and can move the comment out; the current behaviour is silent and is not.

SaFE/charts/primus-safe/values.yaml
```diff
@@ -30,6 +30,40 @@ net:
   rdma_name: "rdma/hca"
   # Ingress component name selected during install.sh
   ingress: "higress"
+node_local_dns:
+  # Forward the queries no more specific zone claims to the CoreDNS of the cluster nodelocaldns runs
+  # in, instead of to the node resolver at /etc/resolv.conf. Enable this to keep site-specific
+  # records in CoreDNS, whose ConfigMap is not reconciled. The address is read from each cluster's
+  # kube-system/kube-dns Service, so one setting stays correct across data-plane clusters.
```
zoroyihan7: **values.yaml still describes the kube-dns lookup 4bdf6ee removed**

The address no longer comes from `kube-system/kube-dns`; the revert paragraph on line 61 repeats the same claim. Dropping the Service name assumption was right, but an operator reading this is told about a mechanism that is not there.

Two properties of the new mechanism are worth stating here, because they are what an operator would get wrong:

Turning this on requires the Corefile to already carry a zone (`cluster.local`, `in-addr.arpa`, `ip6.arpa`) with a `forward .`. Without one the sync errors for that cluster on every reconcile. `guaranteeNodeLocalDNS` is the last step of `Reconcile`, so that cluster then ends every reconcile in error and requeues on error backoff rather than the 15s `controlPlaneEndpointsSyncInterval` cadence — control-plane endpoint probing for that cluster slows down as a side effect of a DNS setting, which is not something anyone would predict from this text.

The revert applies to any root forward holding the cluster DNS address, including one the site wrote (see the comment on `desiredNodeLocalDNSCorefile`). Whatever attribution the code settles on should be described here, since this paragraph is what gets read at the moment someone flips the switch back during an outage.

SaFE/resource-manager/pkg/resource/nodelocaldns.go
```diff
@@ -0,0 +1,334 @@
...
+	for _, server := range cf.Servers {
+		if !servesClusterZone(server) {
+			continue
+		}
+		if plugin := findForward(server); plugin != nil && len(plugin.Args) >= 2 {
+			return plugin.Args[1], true
+		}
+	}
+	return "", false
+}
```
zoroyihan7, line 137: **The first hinted zone wins, and a reverse zone is not always the cluster's DNS**

This returns the forward of whichever hinted zone appears first. In a kubespray file that is `cluster.local` and correct. A site that delegates `in-addr.arpa` to an internal PTR server and declares it first gets that server's address written into the root block instead — an ordinary enterprise layout rather than an exotic one. It does take repointing the whole reverse zone, since `10.in-addr.arpa` does not match the hint.

Cheap tightening: prefer `cluster.local`, and when the hinted zones disagree, error rather than pick. A disagreement means the assumption this reads on — that the cluster zones all name the cluster's own CoreDNS — does not hold for that cluster, and erroring for a single cluster is already the failure mode when no hint is found at all.

zoroyihan7 (commented):
Third pass. The parser swap is the right call, and the reasoning in 4bdf6ee is the reasoning I would have given: every defect so far has been a parsing defect, and hand-rolling a lexer one counterexample at a time produces exactly that sequence. I could not break the block scanning any more — a brace in a comment, a portless root zone, a nested `forward . x {`, an unbalanced file are all handled or refused.

Four of the five round-2 items verified against a harness and resolved.

Two blocking findings, both introduced by the last two commits rather than by the original design, and both the same kind of mistake: deciding what this sync owns by inference instead of by record.

- With `forward_to_cluster_dns` at its default `false`, the sync reverts a root forward the site wrote itself, because it recognises its own writing by address. That reproduces the incident against the exact Corefile shape a site adopts to escape it.
- The serialiser drops comments, and the write guard is byte equality, so the first reconcile after upgrade rewrites a hand-annotated Corefile and deletes them.

Details inline, each with the case I ran.

SaFE/resource-manager/pkg/resource/nodelocaldns.go
```diff
@@ -0,0 +1,334 @@
...
+func revertRootForward(server *corefile.Server) {
+	plugin := findForward(server)
+	if plugin == nil {
+		return
+	}
+	plugin.Args = []string{".", defaultDNSUpstream}
+	if len(plugin.Options) == 1 && plugin.Options[0].Name == forceTCPOption {
+		plugin.Options = nil
+	}
+}
```
zoroyihan7: **force_tcp survives the revert when the site's sub-block carries other options**

A sub-block holding only `force_tcp` is dropped, one holding anything else is kept whole, so a site forward with its own option comes back from the revert as:

```
forward . /etc/resolv.conf {
    max_concurrent 1000
    force_tcp
}
```

That `force_tcp` is this sync's, added by `setRootForward` when the switch went on. The node resolver is now queried over TCP, and an upstream that drops TCP/53 is common enough to be worth avoiding. The revert is the path taken when resolution is already broken, so it is the worst place to leave a setting nobody chose.

Removing the option while keeping the rest of the sub-block matches what the address handling next to it already does. It also falls out for free if ownership becomes explicit, per the comment on `desiredNodeLocalDNSCorefile`.

SaFE/resource-manager/pkg/resource/nodelocaldns.go
zoroyihan7: Verified. Ran the case back through: `# tidy this {` above the file no longer hides the root block, the scan reports one root zone and the stanza joins it instead of a second block being appended, and an unbalanced file is refused outright rather than managed blind. The parser swap in 4bdf6ee retires the depth tracking this was about altogether. Resolving.

SaFE/resource-manager/pkg/resource/nodelocaldns.go
zoroyihan7: Verified: `max_concurrent 1000` survives and gains `force_tcp`, and a second pass adds nothing. One leftover on the way back out, raised separately on `revertRootForward`. Resolving.

SaFE/resource-manager/pkg/resource/cluster_controller.go
zoroyihan7: Verified — there is no Service call left at all, which is a better answer than the one I asked for. Resolving.

SaFE/resource-manager/pkg/resource/cluster_controller.go
zoroyihan7: Verified, the target now reads out of the log without reconstructing it from the delta. Resolving.

SaFE/resource-manager/pkg/resource/cluster_controller.go
zoroyihan7: Leaving this open: the direction is right and the stanza half is fixed, but recognising this sync's own writing by address does not hold, and with the switch at its default the revert now fires on a forward the site wrote itself. Details and the case I ran are in https://github.com/AMD-AGI/Primus-SaFE/pull/713#discussion_r3788383847.

SaFE/resource-manager/pkg/resource/cluster_controller.go
weilei0120: Fixed in 3cc1792. You are right, and the framing that settled it is that address equality is not ownership — the converged Corefile you ran is the shape this incident's own remediation produced, so the default path would have undone the fix the site applied.

Ownership is now recorded, not inferred: `primus-safe.nodelocaldns.forward: cluster-dns` goes on the ConfigMap when this sync points the root forward at the cluster DNS, the revert only fires when it is present, and it is cleared as part of reverting. A forward the site wrote is left alone whatever address it holds. Took the annotation rather than in-memory state for the reason you gave — the claim has to survive a restart of the manager.

`TestGuaranteeNodeLocalDNSKeepsSiteForwardAtTheClusterDNSAddress` is your case: converged Corefile, switch at its default, address matching the cluster zones, and the root forward must come back untouched. `TestGuaranteeNodeLocalDNSRevertsOnlyWhatItRecorded` walks on then off and asserts the annotation appears, the revert happens, and the annotation is cleared so a later run does not revert again. The old test that encoded the address-equality behaviour is gone.

As you noted, this also settles the force_tcp leftover: the revert now drops it whatever else the sub-block carries.

SaFE/resource-manager/pkg/resource/nodelocaldns.go
weilei0120: Fixed in 3cc1792. My doc comment claimed the serialiser writes the file the way it was written, which is not true for comments — the lexer drops them before `renderCorefile` ever runs, and with `desired == corefile` as the only write guard that becomes a silent deletion on the first reconcile after upgrade. The fixture passing because it was comment-free and canonically indented is exactly the gap.

Took the suggestion as written. `unmanageableCorefileReason` now renders what it parsed and compares content, refusing with the first line that would be lost. `droppedLines` returns the lines rather than counting them, and `summarizeCorefileChange` is now two calls to it, so the two share one implementation.

Confirmed the boundary is where you said: your commented example and `forward . x { tls { server_name y } }` are both refused, while tab indentation, CRLF and blank lines between blocks round-trip and stay managed — `TestUnmanageableCorefileToleratesLayout` covers those three so a future change cannot quietly tighten it. Also re-checked the live Corefile from the site: still manageable, still byte-identical on round trip.

Noted in `values.yaml` that such a file is reported and left untouched, and that moving the content out brings it back under management.

SaFE/resource-manager/pkg/resource/nodelocaldns.go
weilei0120: Fixed in 3cc1792. `cluster.local` is now preferred, and when the hinted zones disagree no address is derived. A site delegating its reverse zone to an internal PTR server is ordinary enough that picking whichever zone came first was not defensible.

Covered by `TestClusterDNSAddressPrefersClusterDomain` (reverse zone declared first, cluster domain still wins) and `TestClusterDNSAddressRefusesDisagreement`.

SaFE/resource-manager/pkg/resource/nodelocaldns.go
weilei0120: Fixed in 3cc1792. `revertRootForward` now removes `force_tcp` and keeps the rest of the sub-block, so a site option such as `max_concurrent` survives while the transport this sync chose does not. The point that the revert is the path taken when resolution is already broken is what makes leaving it there the wrong default. `TestRevertRootForwardDropsForceTCPAmongSiteOptions` asserts it.

SaFE/charts/primus-safe/values.yaml
weilei0120: Fixed in 3cc1792. The kube-dns sentence and the revert paragraph both described mechanisms that no longer exist; `values.yaml` now says the address is read from the Corefile's cluster zones and that the revert is driven by the annotation.

Both properties you flagged are documented: that a zone with a `forward .` must already be present, and that disagreeing reverse zones yield no address.

On the backoff side effect, documenting it did not seem like the right answer — a DNS setting slowing control-plane endpoint probing is surprising however well it is written down. A missing or ambiguous address is now reported like the other unmanageable conditions and returns nil, so that cluster keeps the 15s cadence and only its Corefile is left alone.

weilei0120:
Superseded by #715, closing this.

The premise here was wrong. This PR kept the sync and tried to make its edits safe, which means preserving arbitrary content in a file kubespray generates and operators edit — comments, unfamiliar directives, site blocks. That is not a bounded problem, and every review round found another instance of it rather than the same one twice.

#715 removes the sync instead. Resolution is configured in CoreDNS, which nodelocaldns forwards to and caches, so a record changes by editing one ConfigMap and takes effect without rolling nodelocaldns on every node. Nothing in that is dynamic, so no reconcile loop is needed and there is nothing left to preserve.

Thanks for the reviews — each one was reproduced before being acted on, and the last two are what made the pattern legible rather than looking like separate defects.

PR #717, opened 2026-08-17
fix(database): stop the GORM pool serving a demoted replica forever

weilei0120:
## Background

A pooled connection is bound to the backend it was dialled to. On a Postgres cluster that backend can stop being the primary while the connection stays open: the address resolves somewhere else now, but an established connection does not move. Every write on one then fails with `SQLSTATE 25006`, "cannot execute INSERT in a read-only transaction", from a pool that reports itself healthy.

`ConnectGorm` never applied the limits it was handed. GORM opens a pool of its own and this one was left on the `database/sql` defaults, where both `ConnMaxLifetime` and `ConnMaxIdleTime` are zero and a connection is reusable for as long as the process lives. The existing comment observed exactly this and drew only the conclusion that the pool needed its own metrics.

What that produced, on the cluster:

| Time (UTC) | Event |
|---|---|
| 08-13 22:39:00 | `nvsf-0` becomes primary |
| 08-13 22:39:02 | apiserver's GORM pool connects to it — correctly, it is the primary |
| 08-13 22:49:26 | `mtgv-0` takes over as primary; `nvsf` is demoted to a read-only replica |
| 08-13 → 08-17 | two GORM connections stay open on `nvsf`, idle, for 3 days 6 hours |
| 08-17 04:52 | three consecutive `PUT` bindings all draw one of them → 500 |

Requests were served correctly or with a 25006 depending on which connection they drew, which is why it read as intermittent. Every write path on GORM was affected — llm-gateway bindings, image upserts (`image.go:51 "failed to upsert image"`), optimization tasks, models, notifications.

The sqlx pool in the same process, built from the same `DBConfig`, rotated onto the new primary inside its 600s lifetime. That asymmetry — one pool self-healing and the other not, from one config — is what this fixes.

## Changes

**Bound how long a connection may be reused.** Both pools now go through one `applyPoolLimits`, so a second pool cannot be opened without the limits and the two cannot drift again. This takes the exposure from unbounded to one `MaxLifetime`, 600s by default.

A zero still means "leave the default" for the two counts: `database/sql` reads a zero `MaxOpenConns` as unlimited, so forwarding it would uncap a pool the caller meant to leave alone.

**Refuse a session that cannot write.** The lifetime bounds how long a connection to a demoted replica survives; it does not stop one being made. Between a failover and the address catching up, a connection lands on a node that answers, accepts the session, and then refuses every write. `target_session_attrs=read-write` moves that to connect time — the driver asks and declines, so the pool retries rather than handing a caller a connection that will 500. Both drivers support it: `lib/pq` for sqlx, `pgx/v5` under gorm's postgres driver, and both DSNs are built through one helper so they cannot disagree.

Defaulted on, because every component sharing this config writes. Configurable, because it constrains what the secret's address may resolve to.

**Gave that switch a way into a Helm deployment.** It existed only as a config key no template rendered, so the one deployment that must turn it off could not. Read with `hasKey` rather than `default`: an explicitly empty value is how it is turned off, and `default` treats empty as unset and would put `read-write` back — the escape hatch would have rendered as the thing it escapes.

**Checked the error `NewClient` dropped.** A failed GORM init still produced a client, with a nil handle every later query would fault on, arbitrarily far from the boot that caused it. `NewClientWithConfig` has always checked it.

## Verification

- `go build ./common/...`, `go vet`, `gofmt` clean; `common/pkg/database/...` and `common/pkg/config/...` tests pass
- New tests cover both directions, which is what makes them regression tests rather than restatements: a configured lifetime retires a connection and no lifetime never does; a set attr reaches the DSN and an unset one leaves the string byte-identical to before
- Chart rendered three ways: the default asks for `read-write`, `--set db.target_session_attrs=""` forwards the empty, and the Go side honours it because `getString` tests `viper.IsSet` rather than the value

Not covered: that `ConnectGorm` calls the limits. `gorm.Open` pings on open and this repo has no pattern for a test against a live Postgres, so the single shared function is what guards it instead.

`apiserver` does not build in my environment for an unrelated reason — `containers/storage` needs `btrfs/version.h`, which is missing here. `--set db=null` fails to render on `resource-manager/deployment.yaml:62`, which predates this branch and is untouched.

## Deployment

Nothing to do beyond deploying: both defaults are the safe ones. A deployment whose `db` host deliberately resolves to a replica must set `db.target_session_attrs: ""`, or its connections will be refused outright rather than serving reads.

### DIFF
--- a/.github/workflows/block-sensitive-secrets.yml
+++ b/.github/workflows/block-sensitive-secrets.yml
@@ -84,6 +84,14 @@ jobs:
             # DSN builder: "password=%s" is a format placeholder filled from the
             # DBConfig at runtime, not a literal credential value.
             [[ "$path" == SaFE/common/pkg/database/utils/db.go ]] && return 0
+            # Same DSN builders, which now live beside the config they read:
+            # SourceName and GormSourceName both carry "password=%s" as a
+            # placeholder. Listed separately from db.go because the connection
+            # strings moved here to be testable as pure functions.
+            [[ "$path" == SaFE/common/pkg/database/utils/config.go ]] && return 0
+            # Those builders' tests: the expected DSNs contain "password=p", a
+            # single-character stand-in in an assertion, not a credential.
+            [[ "$path" == SaFE/common/pkg/database/utils/dsn_test.go ]] && return 0
             # applyconfiguration-gen output: schema / field names may match generic secret patterns
             [[ "$path" == SaFE/apis/pkg/client/applyconfiguration/* ]] && return 0
             # Frontend form field names and TypeScript type definitions, not credential values
--- a/SaFE/charts/primus-safe/templates/_helpers.tpl
+++ b/SaFE/charts/primus-safe/templates/_helpers.tpl
@@ -1,0 +1,25 @@
+{{/*
+The session a component's database driver must be given.
+
+"read-write" makes the driver check, at connect time, that the session it opened
+can write, and refuse it when it cannot -- so a connection reaching a replica
+fails there instead of failing every write later with "cannot execute INSERT in a
+read-only transaction". Every component reading this writes, so on is the right
+default.
+
+hasKey, not default: an explicitly empty value is how a deployment whose db host
+resolves to a replica turns this off, and `default` reads empty as unset and would
+put "read-write" back -- the escape hatch would render as the thing it escapes.
+The Go side honours the empty because getString tests viper.IsSet rather than the
+value.
+
+Shared rather than repeated in each component's config, because three copies of a
+condition this non-obvious drift apart.
+*/}}
+{{- define "primus-safe.dbTargetSessionAttrs" -}}
+{{- if hasKey (default dict .Values.db) "target_session_attrs" -}}
+{{ (default dict .Values.db).target_session_attrs | quote }}
+{{- else -}}
+"read-write"
+{{- end -}}
+{{- end -}}
--- a/SaFE/charts/primus-safe/templates/apiserver/config.yaml
+++ b/SaFE/charts/primus-safe/templates/apiserver/config.yaml
@@ -50,9 +50,7 @@ data:
       secret_path: {{ default "/etc/secrets/db" (default .Values.db).secret_path }}
       # The ssl mode of db
       ssl_mode: {{ default "require" (default .Values.db).ssl_mode }}
-      # hasKey, not default: an explicitly empty value is the way to turn this
-      # off, and `default` would read it as unset and put the default back.
-      target_session_attrs: {{ if hasKey (default dict .Values.db) "target_session_attrs" }}{{ (default dict .Values.db).target_session_attrs | quote }}{{ else }}"read-write"{{ end }}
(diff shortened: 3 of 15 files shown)


weilei0120:
Thanks — the Medium one is a fair hit, and it was created by my own change: swapping `sync.Once` for a mutex traded "permanent nil" for "retry on every call" and I never bounded the second half. Both paths you named check out — `resource-manager/pkg/health/reporter.go:137` in the periodic self-check, `dispatcher_help.go:185` and `:215` on the reconcile-adjacent path.

Worth adding to the diagnosis: the build holds `instanceMu` across the dial, so during an outage those callers were not just re-dialling, they were serialising behind whoever held the lock. Note that part is not a regression — `once.Do` blocked callers the same way — but repeating it every call is.

**1. Retry cooldown — done** (`2cec704b`). A failure now stands for 30s before another dial is spent on it. Timed from the *start* of the attempt rather than from the failure, so a dial that hangs for its full `connect_timeout` doesn't then earn a further 30s on top; the attempt rate is one per cooldown however slowly the server is failing. Recovery-without-restart is untouched, the cooldown only decides when.

Kept it a package var rather than a config key: the number an operator actually tunes is `connect_timeout`, and this one only has to outlast a single attempt. `TestNewClientHoldsOffAfterAFailure` pins both halves — calls inside the window don't dial, and a failure aged past the window does. I mutation-checked it: neutering the cooldown condition makes it fail with 4 attempts instead of 2, so it's guarding the behaviour and not just passing.

**2. GORM `connect_timeout` — done** (`3692193e`). Took this one rather than deferring it, because it compounds #1: with no timeout on the GORM dial, "how long can one attempt hold the mutex" had no ceiling, so the cooldown alone would have bounded the *frequency* of attempts without bounding their *duration*. It's also exactly the failure mode this PR is about — two strings built from one config, in two places, disagreeing about a parameter — so leaving it as known-asymmetric while fixing the pool limits seemed like the wrong place to stop. The parity test now asserts `connect_timeout` on both DSNs alongside the session attr.

**3. Residual 600s window — agreed, not doing it here.** Actively invalidating the pool on a read-only check is a real improvement and belongs in the health-check path, not in a DSN/pool-limits change.

**4. `MaxLifetime <= 0` warn-not-fail — documented rather than enforced** (`4f8f6e48`). Kept the pass-through, because a caller may mean it and I'd rather not have this function override an explicit config. But your point about operators not reading logs stands, so the note went where the key would be set. Note the key isn't rendered by the chart at all — the 600s default lives in Go — so reaching the unbounded behaviour means deliberately adding an undocumented key, which is partly why warn still seems proportionate.

All of `common/pkg/database/...`, `common/pkg/config/...`, the middleware/email-relay packages and `resource-manager/pkg/health` pass, including under `-race`. The chart change is comment-only — `helm template` output is byte-identical before and after.

Unrelated, flagging so it isn't mistaken for fallout: `TestPlatformKeyForUser/success` in `job-manager/pkg/dispatcher` fails on this branch *and* with my changes stashed. It's a gomonkey inlining issue — passes with `-gcflags=all=-l`. Pre-existing, left alone.

zoroyihan7 (approved):

PR #719, opened 2026-08-20
fix(job-manager): read pod specs at the path the writer used

weilei0120:
## Summary

The drift-detection readers in `job-manager/pkg/utils/unstructured.go` hardcoded a `"spec"` segment after a `ResourceSpec`'s template path, while the writer resolved that segment from the workload kind via `getPodSpec`. The two disagreed for every kind whose slot does not wrap its pod in a `PodTemplateSpec`.

For an `InferaDeployment`, containers live at `spec.services.role0.extraPodSpec.containers`, but the readers looked for `spec.services.role0.extraPodSpec.spec.containers`:

```
failed to find containers, path: [spec services role0 extraPodSpec spec containers]
```

`MonarchMesh` has the same shape: the writer resolves `spec.podTemplate.containers`, the readers looked in `spec.spec.containers`.

## Impact

Every reader on the sync path is affected, not just the entrypoint one — `isEntrypointChanged`, `isImagesChanged`, `isResourceChanged`, `isEnvChanged` and `isSharedMemoryChanged` all resolve containers through the same helper. They do not fail the same way, and that is what makes the bug destructive rather than inert:

- `GetCommands` / `GetImages` return an error that is logged and swallowed into `return false`, and `GetResources` `continue`s past the miss so its result is short and the comparison is skipped — these read as "nothing changed".
- `GetMemoryStorageSize` and `GetPriorityClassName` return an error too, but `isSharedMemoryChanged` and `isPriorityClassChanged` translate an error into `return true` — "everything changed".

The second group wins. For an `InferaDeployment` on `main`, `isSharedMemoryChanged` fires on **every** reconcile, so `applyWorkloadSpecToObject` re-runs every time. The writers resolve their paths correctly, so the re-apply lands — and rewrites the command from `Workload.Spec.EntryPoints` alone. **The sglang flags `normalizeInferaIDEP` injected at create time are stripped on the first reconcile, silently degrading a PD deployment to aggregated**, and a multi-node role's pinned `replicas: 1` is overwritten with the node count, turning one LeaderWorkerSet group of N nodes into N groups.

So the user-visible effect is not a dropped edit; it is a deployment that breaks itself shortly after it is created, with nothing in the log beyond the path error.

`DynamoDeployment` has the same path bug, but `syncWorkloadToObject` returns early for it, so nothing observable changes there. `MonarchMesh` shares the bug in the readers, but `syncWorkloadToObject` is only reached for `IsApplication()` or `IsCICDScalingRunnerSet()` workloads and MonarchMesh is neither, so its paths are fixed here without any behavior change today.

## What changed

- **`podSpecPaths` on `ResourceSpec`**, declared for every template the chart ships. The pod spec location is now data on the CR rather than two independent derivations in code. `ResolvePodSpecPath` prefers the declaration and falls back to the kind-derived default, so an out-of-date `ResourceTemplate` keeps working.
- **`commonworkload.PodSpecSegment` / `BuildPodSpecPath`** — the fallback, extracted from the logic the writer already had. Both the readers and `dispatcher_help.go`'s `getPodSpec` / `buildPodSpecPath` now go through it, so a reader and a writer can no longer drift apart. The resolved paths for existing kinds are unchanged.
- **Path accessors return fresh slices** (`joinPrePaths`). `TemplatePath` used to append onto `PrePaths`' backing array, which could write through and corrupt the cached `ResourceTemplate` for later reads.
- **`InferaDeployment` command handling** — see below.

## Why the InferaDeployment command needs its own handling

Fixing the readers alone would regress Infera. `syncWorkloadToObject` re-derives each command from `Workload.Spec.EntryPoints`, but `normalizeInferaIDEP` appends the sglang disaggregation and multi-node args to the IDEP's launcher payload **without** writing them back to the Workload. With drift detection working, the first reconcile would strip `--disaggregation-mode` / `--disaggregation-transfer-backend` / `--disaggregation-bootstrap-port` and silently degrade a PD deployment to aggregated.

An earlier revision of this branch handled that by exempting the command for `InferaDeployment` as a whole (`d514dbc3`, `10c29afc`). That was too coarse. The injected flags only exist for the roles that get them: `prefill`/`decode` via `applyInferaRoleFields`, and a multi-node role with more than one node via `appendSglangMultinodeArgs`. A `frontend`, or a `worker` that is not multi-node — which is the **default** `[frontend, worker]` shape `GetInferaServiceRoles` falls back to — gets neither, so the exemption dropped a legitimate entrypoint edit for exactly the most common deployment. That is a regression against `main`, where the edit at least failed loudly in the log.

`aa9d78b3` replaces the exemption with a rebuild. `inferaLauncherFlags` reproduces exactly the flags `normalizeInferaIDEP` grafts for a given role, and `expectedCommands` / `expectedEntryPoint` wrap `buildCommands` / `buildEntryPoint` with them, so read side and write side compute the same string. The result:

- an `entryPoints` edit reaches every role, including the flagless ones;
- a steady-state IDEP reports no drift, so there is no reconcile churn;
- the create path is byte-identical — the flags land first and `normalizeInferaIDEP`'s per-flag dedup (`stripUserDeclaredFlags`) makes its second pass a no-op.

Rebuilding `--nnodes` on sync exposed a second half of the same problem: a multi-node role's node count also lives in the slot's flat `numberOfNodes` field, which `normalizeInferaIDEP` wrote on create only. Scaling the role would have moved the flag and left the field behind, giving a LeaderWorkerSet group of the old size a rendezvous for the new one — and since `replicas` stays pinned at 1, no drift check would ever have fired on it. `updateInferaNodeCount` writes the field next to `replicas`, and removes it when the role scales to a single node or leaves `multinode-roles`.

The condition for simplifying all of this later is unchanged: have the injected flags round-trip into the Workload.

## Behavior changes on upgrade

- **`InferaDeployment`**: the reconcile stops rewriting the command from `EntryPoints` alone, so an existing PD or multi-node deployment keeps its injected flags and its `replicas: 1` pinning instead of losing them. An IDEP that `main` has already degraded is repaired on the first reconcile after this rolls out, which restarts its pods — that is the intended repair, but it is a restart.
- **`MonarchMesh`**: paths are corrected, but the readers are never reached for this kind (see Impact), so there is no behavior change.
- **Everything else** (`PyTorchJob`, `Deployment`, `RayJob`, `Job`, …) resolves to the same paths as before and is unaffected.

## Test plan

- [x] Reproduction tests in `job-manager/pkg/utils/unstructured_test.go` over a rendered two-role IDEP fixture, covering commands, images, resources, env and shared memory. All fail on `main` with the path error above.
- [x] `TestShippedResourceTemplatesDeclarePodSpecPaths` parses the real chart, and asserts every shipped `resourceSpec` declares `podSpecPaths` that resolve to the same path as the code default — for the CR's own kind and every kind in its `primus-safe.workload.kind` annotation. A future template that forgets the declaration, or declares a wrong one, fails here.
- [x] Infera command coverage in `dispatcher_help_test.go`: `inferaLauncherFlags` over all role shapes; the `[frontend, worker]` regression (a stale command must be rebuilt); flag re-graft for a multi-node decode role, asserted byte-exact on the decoded payload; create-path idempotency against `normalizeInferaIDEP`; `isEntrypointChanged` for flagless and injected roles.
- [x] `updateInferaNodeCount` covered for scale-up (field and `--nnodes` must agree), scale-down to one node, a role leaving `multinode-roles`, and a non-Infera object left alone — plus a test asserting the sync write lands on the same path `normalizeInferaIDEP` uses on create.
- [x] `appendLauncherFlags` covered for bare, single-quoted and double-quoted payloads, asserting the entry's prefix and trailing quote both survive the splice.
- [x] `expectedReplica` pinned at both production call sites (`isResourceChanged`, `updateReplica`), and `updateMetadata`'s Infera early return covered with a PyTorchJob control.
- [x] Every new assertion mutation-tested — reverting each fix in turn produces exactly the expected failure.
- [x] `go build` and `go vet` clean on the touched modules; `go test ./apis/... ./common/... ./job-manager/... -gcflags=all=-l` fully green. Without `-gcflags=all=-l` two tests fail (`TestPlatformKeyForUser`, `TestServerStartInited`); both are gomonkey patches defeated by inlining, both reproduce identically on a clean `main`, and neither is touched by this branch.

## Notes for the reviewer

- The branch touches 13 files, above the usual limit, but the reader fix and the Infera command handling cannot land separately: shipping the former without the latter is the flag-stripping regression described above.
- `appendLauncherFlags`'s lost trailing quote predates this branch — it was in `appendLauncherArgs` all along. Extracting the string half for the read side is what made it worth fixing here rather than separately.
- `isEntrypointChanged` and its siblings still treat a read error as "unchanged", while `isSharedMemoryChanged` and `isPriorityClassChanged` treat the same error as "changed". That inconsistency is the design flaw that turned a path bug into a self-destructing deployment; making them all surface the error is a larger behavioral change and deliberately out of scope here.

### DIFF
--- a/SaFE/apis/pkg/apis/amd/v1/resource_template_types.go
+++ b/SaFE/apis/pkg/apis/amd/v1/resource_template_types.go
@@ -67,11 +67,16 @@ type ResourceSpec struct {
 }
 
 // TemplatePath returns the path components for locating the resource template.
+// The result is a fresh slice: appending onto PrePaths directly would write
+// through its backing array whenever it has spare capacity, corrupting the
+// ResourceTemplate for later reads.
 func (t *ResourceSpec) TemplatePath() []string {
 	if t == nil {
 		return nil
 	}
-	path := append(t.PrePaths, t.TemplatePaths...)
+	path := make([]string, 0, len(t.PrePaths)+len(t.TemplatePaths))
+	path = append(path, t.PrePaths...)
+	path = append(path, t.TemplatePaths...)
 	return path
 }
 
--- a/SaFE/apis/pkg/apis/amd/v1/resource_template_types.go
+++ b/SaFE/apis/pkg/apis/amd/v1/resource_template_types.go
@@ -85,6 +85,16 @@ func (t *ResourceSpec) joinPrePaths(rel []string) []string {
 	return path
 }
 
+// Path returns the path components for locating a field that sits directly
+// under prePaths and has no dedicated path field of its own. Each call returns
+// a fresh slice, so the result is safe to append to.
+func (t *ResourceSpec) Path(fields ...string) []string {
+	if t == nil {
+		return nil
+	}
+	return t.joinPrePaths(fields)
+}
+
 // TemplatePath returns the path components for locating the resource template.
 func (t *ResourceSpec) TemplatePath() []string {
 	if t == nil {
--- a/SaFE/apis/pkg/apis/amd/v1/resource_template_types.go
+++ b/SaFE/apis/pkg/apis/amd/v1/resource_template_types.go
@@ -56,6 +56,15 @@ type ResourceSpec struct {
 	PrePaths []string `json:"prePaths,omitempty"`
 	// The relative path of pod template
 	TemplatePaths []string `json:"templatePaths,omitempty"`
+	// The relative path of the core/v1 PodSpec, where containers, volumes,
+	// priorityClassName and the other pod-level fields live. Like the other
+	// path fields it is relative to prePaths, not to templatePaths: a kind that
+	// wraps its pod in a PodTemplateSpec repeats the template segment and adds
+	// "spec" (template, spec), while a kind whose slot embeds a PodSpec inline
+	// names that field alone (extraPodSpec).
+	//
+	// Empty means the path is derived from templatePaths and the kind instead.
+	PodSpecPaths []string `json:"podSpecPaths,omitempty"`
 	// The relative path of pod replica
 	ReplicasPaths []string `json:"replicasPaths,omitempty"`
 	// Fits scenarios without a defined replica path but with actual replica values, such as ray-job.
@@ -82,6 +91,15 @@ func (t *ResourceSpec) TemplatePath() []string {
 		return nil
 	}
 	return t.joinPrePaths(t.TemplatePaths)
+}
+
+// PodSpecPath returns the path components for locating the pod spec, or nil
+// when the resource spec does not declare podSpecPaths.
+func (t *ResourceSpec) PodSpecPath() []string {
+	if t == nil || len(t.PodSpecPaths) == 0 {
+		return nil
+	}
+	return t.joinPrePaths(t.PodSpecPaths)
 }
 
 // ReplicasPath returns the path components for locating the replica count.
--- a/SaFE/apis/pkg/apis/amd/v1/resource_template_types.go
+++ b/SaFE/apis/pkg/apis/amd/v1/resource_template_types.go
@@ -66,18 +66,46 @@ type ResourceSpec struct {
 	MinReplicasPaths []string `json:"minReplicasPaths,omitempty"`
 }
 
+// joinPrePaths returns PrePaths followed by rel as a fresh slice, leaving
+// PrePaths' backing array untouched even when callers append to the result.
+// The path accessors below all build on it.
+func (t *ResourceSpec) joinPrePaths(rel []string) []string {
+	path := make([]string, 0, len(t.PrePaths)+len(rel))
+	path = append(path, t.PrePaths...)
+	path = append(path, rel...)
+	return path
+}
+
 // TemplatePath returns the path components for locating the resource template.
-// The result is a fresh slice: appending onto PrePaths directly would write
-// through its backing array whenever it has spare capacity, corrupting the
-// ResourceTemplate for later reads.
 func (t *ResourceSpec) TemplatePath() []string {
 	if t == nil {
 		return nil
 	}
-	path := make([]string, 0, len(t.PrePaths)+len(t.TemplatePaths))
-	path = append(path, t.PrePaths...)
-	path = append(path, t.TemplatePaths...)
-	return path
+	return t.joinPrePaths(t.TemplatePaths)
+}
+
+// ReplicasPath returns the path components for locating the replica count.
+func (t *ResourceSpec) ReplicasPath() []string {
+	if t == nil {
+		return nil
+	}
+	return t.joinPrePaths(t.ReplicasPaths)
+}
+
+// MaxReplicasPath returns the path components for locating the max replica count.
+func (t *ResourceSpec) MaxReplicasPath() []string {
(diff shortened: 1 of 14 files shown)


weilei0120:
Thanks — all three findings are addressed in `50911dbd`, plus the two factual errors in the PR description, which I have rewritten.

**`dispatcher_help.go:767` — `numberOfNodes` left behind when `--nnodes` moves.** Confirmed, and it is a regression `aa9d78b3` introduced: before that commit the flag was excluded from the sync, so the field and the flag were both frozen at their create-time values and agreed with each other. Rebuilding the command made the flag move while the field stayed, which is exactly the state the `expectedReplica` comment assumed could not happen — that comment is updated too.

Fixed by syncing the field: `updateInferaNodeCount` writes it next to `replicas` in `updateReplica`. One addition beyond the report — scaling down needs the field **removed**, not set to 1. `normalizeInferaIDEP` does not write `numberOfNodes` at all for a single-node role, so a 4→1 scale that only ever writes would leave `numberOfNodes: 4` next to `replicas: 1` and break the same way the scale-up case does. Same for a role dropping out of `multinode-roles`. `TestInferaNodeCountMatchesCreatePath` now asserts the sync path and the create path produce the same shape.

**`resourcespec.go` — missing `PodSpecPaths`.** Regenerated in `8c1fd535`.

**`dispatcher_help.go:1817` — payload splice drops the tail.** Fixed as suggested, `full[:at] + newPayload + full[at+len(payload):]`, with a guard for `LastIndex` returning -1 (the old code would have panicked on the slice). Covered for bare, single-quoted and double-quoted payloads. Agreed it predates this branch; extracting the string half for the read side is what made it worth fixing here.

**PR description.** Both errors were real and I have verified them independently:

- The "sync has never done anything" framing was wrong. `isSharedMemoryChanged` turns the read error into `return true`, so `main` re-applies an IDEP on every reconcile, and since the writers resolve correctly the re-apply lands — stripping the injected sglang flags and overwriting the pinned `replicas: 1`. The description now says that an IDEP on `main` breaks itself shortly after create, and spells out the two reader groups whose opposite error handling causes it.
- The MonarchMesh upgrade warning was wrong. `syncWorkloadToObject` is only reached for `IsApplication()` or `IsCICDScalingRunnerSet()`, MonarchMesh is neither, and those five readers have no other callers. Removed, and the section now says what actually happens on upgrade: an IDEP already degraded by `main` is repaired on the first reconcile, and that repair restarts its pods.

**Not changed, flagging for your call:** an external edit of `numberOfNodes` directly on the object is still not detected — the command is unchanged, so `isEntrypointChanged` does not fire. The scale path is covered because `--nnodes` moves with it. I left this alone because the other Infera-only flat fields (`componentType`, `role`, `podLabels`) have no drift check either, so singling out `numberOfNodes` would be inconsistent. Happy to add all of them if you would rather close that gap here.

Every new assertion is mutation-tested. `go test ./apis/... ./common/... ./job-manager/... -gcflags=all=-l` is green.

zoroyihan7 (approved):

PR #721, opened 2026-08-25
fix(resource-manager): keep a node bound to a single workspace

weilei0120:
## Problem

A node could be bound to two workspaces at once, and the workspace that lost the race was left wedged for the lifetime of the process — no scaling, no nodes actions, no status sync — until the leader was restarted.

Observed in the field: `crsuse2-m2m-237` was bound to `control-plane-hyperloom`, then taken over by `control-plane-spur-quarantine` a few hundred milliseconds later. `hyperloom` never reconciled again, and a later request to bind `crsuse2-m2m-172` to it sat unprocessed.

Two independent defects combine to produce that:

1. **The binding patch carries no precondition.** `client.MergeFrom` produces a JSON merge patch with no resourceVersion, so two binders that each read the same free node from their caches both succeed and the last writer wins. No error, no conflict, no signal.

2. **Expectations are credited by the wrong key.** A workspace registers an expectation on every node it is binding and refuses to reconcile until each one is observed. The observation was credited by the workspace id the node event carried, so once a node changed hands the workspace actually waiting on it was never credited. Its expectation stayed in the map and `processWorkspace` returned at the `meetExpectations` gate on every subsequent reconcile.

## Changes

**Refuse the second binder** — `updateSingleNodeBinding` rejects a bind whose target node is already owned by another workspace (it has to be unbound there first), and patches through `client.MergeFromWithOptimisticLock` so a concurrent binder that got there first produces a conflict instead of a silent overwrite. The nodes-action path drops such a request rather than retrying a binding that must not happen, so the annotation clears.

**Release the expectation by node, not by owner** — `observeNodeForAll` settles a node against every workspace waiting on it and re-queues each one so it can pick a different node. Both the refusal and the conflict path already run through `observeNode` in `updateNodesBinding`, so a failed bind no longer leaves a dangling expectation. No TTL and no `RequeueAfter` are involved.

**Let explicit bindings win over automatic scaling** — scale-up skips nodes another workspace is already processing, whether the binding is in flight or still sitting in an unprocessed nodes action.

**Do not accept a binding that can never settle** — a binding is only settled once the workspace label has made the round trip through the data plane. `updateK8sNode` returns early when the k8s Node is missing, so binding a node that has left the cluster would wait on a label that never arrives. The add path now checks the node is still there and drops the request otherwise. A data plane that fails to answer counts as present, since refusing here is permanent. Only the add path needs this — when a k8s Node disappears, `handleNodeUnmanaged` clears its admin labels and settles any unbind on its own.

**Stop the data plane from writing ownership** — the reverse metadata sync copied the workspace label off the k8s Node unconditionally, making the data plane a third writer of an ownership the admin plane is supposed to decide. A value that disagrees with `spec.workspace` is stale or was edited by hand, and copying it back makes the admin plane report an ownership nobody asked for — to every label selector, and to the ops-job webhook that validates node membership by it. Those values are now ignored and `NodeReconciler` pushes the right one back down. Unbinding is unaffected: the label still goes away through the removal branch once the data plane has dropped it.

## Behaviour change

Binding can now fail where it previously always "succeeded". This only happens in the contention case that used to corrupt ownership silently:

- ownership pre-check → `(false, nil)`, not an error: a warning, a `bind/conflict` metric, and the request is dropped.
- optimistic-lock conflict → error, propagated to `Reconcile` for the standard rate-limited requeue, re-evaluated against fresh state.

Uncontended binds are unchanged.

## Tests

12 new cases in `workspace_node_ownership_test.go`, including a regression for the exact `crsuse2-m2m-237` sequence and for a stale write being rejected by the optimistic lock. `TestWorkspaceNodesAction` and `TestSyncK8sMetadata` were adjusted for the new signature and still pass.

`go build ./resource-manager/...` is clean; `go vet` shows only the pre-existing `addon_helper.go:84` warning. Five failures remain in `resource-manager/pkg/resource` (`TestGuaranteeCICDClusterRoleBindingFull`, `TestGuaranteeForwardIngressFull`, `TestConstructDownloadJobFull`, `TestConstructCleanupJobFull`, `TestConstructLocalDownloadOpsJob`) — reproduced identically on a clean tree via `git stash -u`, unrelated to this change.

## Not included

Tightening `node_webhook.validateImmutableFields` to reject an `"" → X` ownership jump. With the reverse sync no longer writing ownership it is largely a redundant second line of defence; worth revisiting after this has run for a while.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/SaFE/charts/primus-safe/templates/configmap/stateful_set_template.yaml
+++ b/SaFE/charts/primus-safe/templates/configmap/stateful_set_template.yaml
@@ -62,7 +62,7 @@ data:
                - name: NCCL_CHECKS_DISABLE
                  value: "1"
                - name: NCCL_IB_SL
-                 value: "0
+                 value: "0"
                - name: POD_IP
                  valueFrom:
                    fieldRef:
--- a/SaFE/resource-manager/pkg/resource/node_controller.go
+++ b/SaFE/resource-manager/pkg/resource/node_controller.go
@@ -176,6 +176,11 @@ func (r *NodeReconciler) delete(ctx context.Context, adminNode *v1.Node) (ctrlru
 }
 
 // getK8sNode retrieves the Kubernetes Node object in the data plane associated with the admin Node.
+//
+// Deliberately the uncached read: what this returns is written back -- Nodes().Update,
+// Nodes().UpdateStatus, and a patch body carrying its resourceVersion. kubelet moves that
+// version every few seconds with its heartbeat, so an informer copy would make all three
+// conflict periodically and drop taint/label/condition syncing into backoff.
 func (r *NodeReconciler) getK8sNode(ctx context.Context, adminNode *v1.Node) (*corev1.Node, ctrlruntime.Result, error) {
 	clusterName := getClusterId(adminNode)
 	k8sNodeName := adminNode.GetK8sNodeName()
@@ -186,7 +191,7 @@ func (r *NodeReconciler) getK8sNode(ctx context.Context, adminNode *v1.Node) (*c
 	if err != nil || !k8sClients.IsValid() {
 		return nil, ctrlruntime.Result{RequeueAfter: time.Second}, nil
 	}
-	k8sNode, err := getNodeByInformer(ctx, k8sClients, k8sNodeName)
+	k8sNode, err := getDataPlaneNode(ctx, k8sClients, k8sNodeName)
 	return k8sNode, ctrlruntime.Result{}, err
 }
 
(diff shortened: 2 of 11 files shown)


weilei0120:
Superseded by #723, which rebuilds this from scratch on a different principle: refuse before the danger, at admission where the caller sees it, and never silently drop a request in the controller.

That removes the machinery this PR needed to make its dropped requests safe — the event recorder, `k8sNodeUnavailableReason`, the informer/apiserver read split, the workspace-by-cluster field index — and closes the two findings raised here (the conflict re-read decoding into a mutated object, and the dropped action stranding the `Spec.Replica` the mutating webhook already applied). Production diff goes from +581 across 11 files to +272 across 7.

Not deleting the branch, in case anything here is worth going back to.

PR #723, opened 2026-08-26
fix(resource-manager): make node ownership a compare-and-swap

weilei0120:
Replaces #721, which took the opposite approach and did not converge.

## The problem

A node belongs to at most one workspace, and only its owner may release it. Nothing enforced that at the write. `spec.workspace` was patched from a cached copy with no ownership check at all, so:

- two workspaces reaching for the same free node both succeeded, and the second silently overwrote the first — both then believed they owned it;
- posting `{"node":"remove"}` for somebody else's node was a supported way to take it away from them. The rule was open-coded in three places, in three shapes, and every one of them keyed on a non-empty target — an unbind's target is always empty, so it slipped past all three.

## The fix

`updateSingleNodeBinding` is now the only place ownership is written, and it is a compare-and-swap:

```
apiReader reads into a fresh object  →  judgeNodeBinding  →  patch under an optimistic lock
                    ↑                                                      │
                    └──────────────── conflict: read again, judge again ───┘
```

Two workspaces racing for one free node: both pass admission (at that moment the node really is unowned), both read it free. A's patch carries resourceVersion R and lands; B's still carries R and is rejected as a 409, so B re-reads, sees the node is A's, and refuses. The judgement and the write cannot be separated, so nothing fits between them.

`judgeNodeBinding` is that rule as one predicate instead of three that disagreed.

Three layers, in the order a request meets them:

1. **Admission** — `validateNodesAction` refuses what it can see, synchronously, where the caller gets the answer.
2. **`reservedNodes`** — automatic scale-up leaves alone any node another workspace has already asked for by name. An explicit request is something a user made and admission accepted; letting a scale-up take the node out from under it turns that into a failure the user cannot do anything about.
3. **The CAS** — the last line, for the window between 1 and 3.

## Refusals are errors, not silent drops

The nodes-action annotation is cleared only once the whole batch succeeds, and the mutating webhook has already applied the matching `Spec.Replica` change in the same admission request. Dropping a request in the controller would strand that replica and quietly turn a refused explicit bind into an automatic scale-up onto some other node. Returning an error keeps the request visible and the two in step.

The trade-off, stated plainly: a genuine TOCTOU loss leaves the annotation in place, which blocks further node operations on that workspace until someone clears it (setting the annotation to `{}` is accepted). That is a visible, recoverable state, unlike a stranded replica.

## Two supporting fixes

**`observeNodeForAll`** settles a node against every workspace waiting on it, not just the one the node currently points at. An expectation waits for the workspace label to make the round trip through the data plane, and the event only carries the node's new or empty workspace id — not always the id that is waiting. A bind that landed and was followed by the node's deletion left the binding workspace waiting forever, and every later reconcile of it returned early on `meetExpectations`, including the one that would have deleted it. For the same reason a node whose label already reads the target settles immediately: no admin Node write is coming to carry a transition that has already happened.

**`syncK8sMetadata`** no longer copies a data-plane workspace label that disagrees with `spec.workspace`. Ownership is decided in the admin plane; that label is only the confirmation the binding landed, so a value that disagrees is stale or hand-edited, and copying it back would make the admin plane report an ownership nobody asked for.

## Webhook

A node still being onboarded has no cluster label yet — the write in `NodeReconciler.manage` that stamps `ClusterIdLabel` is the same one that sets `Managed` — so adding one to a workspace failed the *cluster* check and reported a mismatch that does not exist. Checked ahead of the cluster check, and for adds only: a node that ended up bound and then lost its managed state is exactly the one that needs releasing.

## Also

`fix(charts)`: `NCCL_IB_SL` opened a double quote and never closed it, so YAML kept scanning into the next line and swallowed the `POD_IP` entry after it.

## Testing

`go build` passes on each commit for both modules. Webhook suite green. resource-manager compared commit-by-commit against `main`: identical failure set, all pre-existing (S3 not enabled, etc.). `TestConstructLocalDownloadOpsJob` panics on `main` and takes the test binary with it, so the comparison skips it — otherwise the results are not meaningful.

22 new tests. Two of them were mutation-checked to confirm they are load-bearing:

- dropping the optimistic lock → the two-workspace race test fails, ws1 overwrites ws2;
- reading into the caller's object instead of a fresh one → that regression test fails. (`client.Get` decodes without zeroing the target, and `Spec.Workspace` is an `omitempty` pointer, so a retry would judge its own previous write as settled. The fake client zeroes, so this only appears against a real apiserver.)

Also covered: refusal returns an error with the annotation left in place, the deleting-node bind/unbind split, both expectation wedges, and the scale-up reservation.


weilei0120:
Follow-up commit addressing the review. Summary of what changed and what I pushed back on.

### 1. The wedged annotation — fixed, but the mechanism is not the one described

The chain through `reservedNodes` does not actually cost other workspaces anything. `getNodesForScalingUp` already drops any node that is owned, unmanaged, or not machine-ready before it consults `reserved` (`workspace_controller.go:549-552`), and every permanent refusal is about a node in exactly one of those states. A dead claim on `n1` reserves a node that was already invisible to B/C/D. The reservation is redundant there, not harmful.

What the wedge does cost is the workspace that made the request. `processWorkspace` returns at the `processNodesAction` error and never reaches `syncWorkspace`, so ws1 stops scaling entirely — with `Spec.Replica` sitting +1 above a node it never got — until somebody hand-edits the annotation to `{}`. That is worse than the externality, and it is the reason I took the stronger fix rather than the documentation one.

`processNodesAction` now judges each entry against its own uncached read and withdraws the ones that cannot be applied: out of the annotation, `Spec.Replica` given back what the mutating webhook added for that entry, reason written to a new `WorkspaceNodesActionError` annotation on the Workspace. One patch for both edits, so a failure between them cannot roll the replica back twice — and the rewritten annotation is what makes the second pass a no-op. An entry that is merely already true is not a refusal and is not charged, matching the webhook, which skips those when it counts.

`updateSingleNodeBinding` still returns a refusal as an error. It is now a one-round detour rather than a loop: the next `processNodesAction` reaches the same verdict against its own read and withdraws.

### 2. Premature settling — fixed, but not with the suggested rule

`GetWorkspaceId(newNode) == "" || == W` still settles ws2 in the handoff case, because the intermediate event in `ws1 → "" → ws2` carries exactly the empty string. Settling on old-or-new instead — which is what base did — leaves ws2 alone and still settles ws1's unbind. Delete keeps `observeNodeForAll`, since no further event is coming for a node that is gone; that was the case the widening was for.

### 3. `cleanupNodeAfterUnmanage` — both halves

Comment corrected, and the path had a real leak: it released `spec.workspace` only when the node also carried the workspace label, so a node unmanaged inside a bind's round trip kept the claim permanently — and nothing could clear it, since only the owner may release and the owner has been told the node is gone. It now releases on the claim rather than the label. Left deliberately outside the ownership rule and deliberately not touching `Spec.Replica`, both now written down there.

### 应修

- Managed check at the write layer: folded into `judgeNodeBinding`, along with the deleting-node check that was sitting loose above it. One function, every reason, `*v1.Node` in.
- `validateNodesAction` on unrelated updates: returns early when the actions map is unchanged. After the existing "another job is processing" check, reaching the loop means the request is new.
- The `syncK8sMetadata` warning: `V(4)`.
- `judgeNodeBinding` ignoring `requester` on the bind path: added.

### 覆盖

Assertions added for the `refused` and `conflict` label values, for `reservedNodes`' error reaching the caller, and for the read-failure counter (which is new — a non-NotFound `apiReader.Get` failure emitted nothing, so throttling read as traffic going to zero rather than as failure). Five mutations confirmed the new tests are load-bearing: reverting the observe tightening, dropping the replica rollback, restoring the refuse-as-error path, tying the unmanage release back to the label, and swallowing the `reservedNodes` error each turn exactly the intended test red.

### 小项

Third copy of the annotation parser removed — `reservedNodes` and `processNodesAction` share one, which also settles the disagreement about parse failures. `reservedNodes` moved after candidate filtering. Webhook message now reports both halves of `IsManaged`.

### On the `delete()` finding

Agreed, and it holds for the same reason after this change.

Both modules build; `resource-manager`'s failure set is byte-identical to base (7 pre-existing, `TestConstructLocalDownloadOpsJob` skipped on both sides as it nil-derefs without S3); `webhooks` is green.

zoroyihan7 (approved):

PR #726, opened 2026-08-27
feat(apiserver): ssh -R remote port forwarding with a pod-internal listener

zoroyihan7:
## What

Implements standard SSH remote port forwarding (`ssh -R` / `RemoteForward` /
`tcpip-forward`) in the SaFE SSH gateway, with the listen socket created **inside
the user's target pod**.

Until now the gateway discarded every global request, so a client asking for a
remote forward got `Warning: remote port forwarding failed for listen port ...`
and Cursor's `RemoteForward` silently did nothing.

The motivating case: pod-resident tools (`git`, `gh`, `pip`) that need to reach
services which only allowlist a developer's own network, by routing through a
proxy on their machine.

```
pod curl 127.0.0.1:10800 -> socat listener in the pod -> K8s exec
  -> apiserver -> forwarded-tcpip -> SSH -> developer's machine
```

On by default (`ssh.reverse_forward.enable`), with loopback-only bind, a bounded
port range, a per-session forward limit, and authorization against the workload on
every request. A deployment that does not want pod traffic leaving through a
developer's network sets it to false.

## Why listen inside the pod

For an authoring workload the pod is the dev box the user thinks they are on,
which is where OpenSSH would put the socket. An apiserver-side listener would
share one port space between tenants, and would not be reachable at
`127.0.0.1` from inside the pod — which is the address users actually configure.

## Commits

| Commit | Contents |
|---|---|
| `feat(config)` | Policy settings, chart values and template, plus a test that renders the chart through `LoadConfig` so template and getters cannot drift |
| `feat(apiserver)` | Global request handling, the forward manager, and the pod-side relay |
| `docs` | User-facing `-L` / `-R` section with the requirements and limits |

Each commit builds and passes its tests on its own.

## Two shell details worth knowing before editing the relay

Both cost real debugging time and both read as correct in a diff:

- **A shell gives a background job `/dev/null` for stdin.** Both relay scripts dup
  their own stdin onto fd 3 before backgrounding socat. Without it the accepted
  socket is never read and the pod-to-client direction carries an immediate EOF
  instead of data — the feature looks wired up and moves no bytes.
- **`set -e` stays in force inside a trap.** A `kill` of an already-exited process
  ends the trap before it removes the rendezvous directory. Both cleanup routes
  guard that kill with `|| true`.

## Teardown

Teardown deliberately does not rely on the trap: the container runtime kills the
exec'd process outright and a trap never runs. Three independent routes:

1. Ending the exec's stdin — the primary signal, and the only one a runtime that
   leaves the process running will honour.
2. A detached `/proc` watcher that reaps the listener if the script owning it
   disappears (covers SIGKILL).
3. A per-connection watcher that bounds a child relay to the life of the forward
   that owns it, so nothing survives into the pod's own lifetime.

Route 3 is worth stating precisely, because it is easy to read as more than it is:
it watches the forward's rendezvous directory, not its own connection, so it
reaps at teardown rather than the moment one connection is abandoned. What bounds
an individual connection is the bridge — when the channel closes, the exec stream
goes with it and the pod-side pair follows. Observed directly during testing: an
hour of channels being rejected by the client left zero relay processes behind.

Every route out of a forward passes through `closeForward`, so an `established`
line in the audit trail always has one matching close with a reason and duration.

## Testing

Unit and integration tests run offline; every new production line was
mutation-checked (deleting it fails a test). Clean under `-race`.

Validated end to end against a live cluster with this image deployed:

- `ssh -N -R 127.0.0.1:10800:127.0.0.1:7890` accepted, no failure warning
- `ss -lntp` shows the listener on pod loopback
- `curl 127.0.0.1:10800` from inside the pod served by the developer's machine
- 8 concurrent requests all 200; 276 KB transferred twice, byte-identical
- 11 connections left zero leftover processes and no leftover sockets
- On disconnect: listener gone, relay processes gone, rendezvous directory gone,
  port rebindable

Two gated test files carry the layers offline tests cannot reach (Kubernetes exec
transport against a real pod; helm template through the config getters). Both
skip unless a cluster / helm is available, so CI is unaffected.

## Reviewed and deliberately not changed

Two review findings were checked and left alone; recording them so the next reader
does not have to re-derive the same answer.

**`dial` failing per connection.** `dial` has one error return, from
`NewSPDYExecutor`, which fails at the rest-config/TLS level. That config is the one
the acceptor exec already used, so if it were broken the forward would never have
been established — the path is not reachable per connection. The `continue` stays
as a guard. The failure shapes that *are* reachable already self-heal: a channel
the client rejects closes the exec stream and the pod-side pair with it, and
announcements still buffered at teardown are reaped with the rendezvous directory.

**A connection announced under a returning id.** Ids are now claimed with `mkdir`,
which is atomic, so two children cannot hold one. The residual case — the apiserver
attaching to a stale announcement after a full PID cycle — would need it to be that
far behind; measured `pid_max` on the target nodes is 4194304.

## Known follow-ups

- Global requests are answered one at a time and `handleForward` does its work
  inside that loop, so a slow forward — an unreachable cluster, a pod that will not
  bind — delays the requests behind it. x/crypto/ssh buffers 16 incoming requests,
  and past that the connection's other channels stall with them. Draining the queue
  into the manager would decouple them; answering out of order would not, because
  the reply carries no request id and the client pairs replies by arrival.
- Authorization runs before the per-session reservation, so a client re-asking for a
  port it already holds still drives a full round of control-plane calls. Reserving
  first, and releasing on the refusal paths, would put that behind the limit that
  exists for it.
- A listener whose exec slots are all busy backpressures into the same queue that
  carries its diagnostics, so `SAFE-RFWD-ERR` and `SAFE-RFWD-DROP` stall with it.
- The script's readiness deadline starts when the exec stream is up, the
  apiserver's when the exec is requested. An exec that takes more than the margin
  between them puts the apiserver's generic timeout in front of the script's
  specific reason.

- Global requests are answered in order, and one slow forward delays the ones
  behind it. Draining the request queue would decouple them; answering out of
  order would not, because the reply carries no request id and the client pairs
  replies with requests by arrival. The resolve and readiness timeouts are sized so
  the worst case stays inside what a client waits before giving up.
- The per-connection exec goroutines are not tracked by the session's WaitGroup, so
  `closeAll` can return while their streams are still unwinding.
- One SPDY exec per TCP connection is the wrong cost model for something like
  `pip install` behind `HTTPS_PROXY`. This is the tradeoff the bundled relay binary
  below is meant to close.

## Rollout

Enabled on merge. A deployment can turn it off with
`ssh.reverse_forward.enable: false`. The MVP requires `socat` in the workload image
— present in the authoring images — and a forward is refused with a clear reason
where it is missing; a bundled relay binary is left for a follow-up.

Plan: `.cursor/plans/ssh-reverse-port-forward-pod-listen.md`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


weilei0120 (approved):

weilei0120:
@copilot resolve the merge conflicts in this pull request

PR #750, opened 2026-09-15
Keep kubespray upgrade pods off the drain path and add kubeletMaxPods

weilei0120:
## Summary
- Skip SaFE upgrade worker pods during Kubespray drain so in-cluster upgrades are not evicted by their own playbook.
- Add optional `kubeletMaxPods` on Cluster control plane, pass it to Kubespray, and expose it on create, upgrade, and cluster detail (Web default 110).

## Test plan
- [ ] Create a cluster from the UI and confirm `kubeletMaxPods` defaults to 110 and is written to the Cluster spec.
- [ ] Upgrade a cluster and confirm the upgrade Pod command includes `drain_pod_selector=primus-safe.cluster.manage.action!=upgrade`.
- [ ] Change only `kubeletMaxPods` on a Ready cluster and confirm a reconcile/upgrade run applies `-e kubelet_max_pods=<n>` without requiring an image/version change.


siliangchen-amd (approved):
Approved.

